Issue #3, opened 2020-08-03
Schedule first TAC meeting


caniszczyk:
We should setup a simple doodle pool, I believe the third week of August was floated due to people's vacations.

cc: @rhaning @dlorenc

dlorenc:
Doodle poll here: https://doodle.com/poll/yya9ai9txhi55wct

Please fill out if you're interested in attending.

caniszczyk:
It's on the public calendar, it's August 21st

https://calendar.google.com/calendar?cid=czYzdm9lZmhwNWk5cGZsdGI1cTY3bmdwZXNAZ3JvdXAuY2FsZW5kYXIuZ29vZ2xlLmNvbQ

PR #4, opened 2020-08-03
Update README.md

svarughese91:
added myself as a member


dlorenc:
Hi @svarughese91!

We won't be able to merge this. The initial TAC membership is defined in the charter over here: https://github.com/ossf/foundation/blob/main/OpenSSF%20Participation%20Agreement%20and%20Funding%20Charter%2020200716.pdf

The TAC meetings are getting scheduled now, and attendance is open, so we'd still love to have you participate! Follow along at https://github.com/ossf/tac/issues/3 for scheduling.

Issue #14, opened 2020-08-21
update README with how to engage


christo4ferris:
Should update the README to make clear:
- TAC meetings open to anyone
- Use GH Issues to request/discuss agenda items
- link to calendar and indicate recurring meeting time/zoom link

LindsayLF:
@christo4ferris I've completed the pull request to make these updates.

estesp:
seems like this issue is complete? @christo4ferris

rhaning:
TAC agreed that this is complete now. Closing issue.

PR #27, opened 2020-10-01
Add slack and github issues links

dcmiddle:
Addresses _issues_ portion of https://github.com/ossf/tac/issues/14
and adds slack.

The `Get Involved` section heading comes from the proposed template https://github.com/ossf/project-template/pull/1

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,13 +2,22 @@
 
 The OpenSSF Technical Advisory Council is responsible for oversight of the various Technical Initiatives of the OpenSSF.
 
+
+## Get Involved
+
+Official communications occur on the [TAC mailing list](https://lists.openssf.org/g/openssf-tac/topics). [Manage your subscriptions to Open SSF mailing lists](https://lists.openssf.org/g/main/subgroups).
+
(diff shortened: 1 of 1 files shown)


mayakacz (approved):

christo4ferris (approved):
LGTM

christo4ferris (changes requested):
actually, checking the slack link and it appears it has expired. Will need to update with a valid invite link

LindsayLF:
Hi @christo4ferris - Here is a new invite link for slack: https://join.slack.com/t/openssf/shared_invite/zt-hxttbilt-R4bR7Gcp0eY2E8zZZZD65w 
(They expire approx every 30 days)

dcmiddle:
Updated with a fresh link. As Lindsay mentions it looks like these expire after a month. We'll want to come up with a more permanent way to link this. From a brief search it seems one needs to setup some ad-hoc mechanism to workaround slack's limitation. To make sure we can ramp people into our community during this initial phase, though, I think it still makes sense to post the ephemeral link and not wait for creating new infrastructure.

Issue #13, opened 2020-08-21
Create lifecycle for WGs + maturity levels


caniszczyk:

dlorenc:
I thought through this a little bit more. Here's what I think we should be trying to achieve/prevent with this process.

The TAC exists to provide oversight of initiatives. What I think we want to avoid is a bunch of working groups taking part in overlapping, potentially conflicting activities that result in unclear guidance. So, to address this, I propose we basically have two tiers of Technical Initiatives:

### Incubating

This includes all existing WGs, and new proposed ones.
To form a new incubating Initiative, we probably want some of the following:

- 1-2 TAC sponsors
- Rough idea for goals/scope
- Enough evidence of interest (multiple companies, 5+ people committed to attend)
- Participate in quarterly reviews

These initiatives:

- May operate as part of the OSSF
- Have access to community resources (Zooms, Youtube channels, GitHub, Slack channels, etc.)
- Can request funding/other resources (subject to TAC/GB approval)
- Can publish things under the OSSF banner, with further TAC approval. The TAC will review all formal publications, launched services, etc. to ensure they fit in with the rest of the OSSF efforts, and represent the technical vision of the TAC.

#### Graduated Initiatives

These meet all the criteria of an incubating Initiative, plus:

- A formal charter reviewed by the TAC annually
- Formal governance, with multiple organizations represented

These can:

- Publish things and launch services without further TAC review, provided these are within the scope of the approved charter.

kaywilliams:
I have a couple items to consider adding. Notes - 1) I am keen on the idea of asking initiatives (both incubating and graduated) to set quarterly goals and provide quarterly progress summaries so that for the OpenSSF overall we can demonstrate momentum and celebrate success. 2) I would like to guide initiatives to have consistency across readme and governance documents.  This helps people who want to participate know where and how to get involved.  I am open to thoughts from others.

Incubating
* Draft readme.md and governance/charter.md for approval by TAC.
* Establish quarterly goals and provide quarterly progress summary to support OpenSSF communications (quarterly press releases)

Requirement for Graduated
* Establish quarterly goals and provide quarterly progress summary to support OpenSSF communications (quarterly press releases)

dlorenc:
This sounds great!

dlorenc:
At second thought, and after looking at some other OSS projects I've been involved in, quarterly goals seem a little too granular. Something like six months or even a year is more realistic for goals.

Quarterly status updates and updates to those goals still make sense though.

[...]

PR #41, opened 2020-12-02
Add Working Group Life Cycle

dcmiddle:
The initial draft of the working group life cycle is based on
https://github.com/ossf/tac/issues/13 and TAC discussions leading up to
the 2020-12-01 meeting.

It is intended to satisfy the Open SSF Charter item 6.e.iii:
[The TAC will be responsible for: ...]
"iii) creating, maintaining and amending project lifecycle states, review
procedures and processes;"

Signed-off-by: Dan Middleton <dan.middleton@intel.com>

### DIFF
--- a/working-group-lifecycle.md
+++ b/working-group-lifecycle.md
@@ -45,8 +45,10 @@ progress a project will move to `End-of-Life`. The details for each phase follow
 
 ## To become `Active`:
 
-* Have met at least 20 times
-* Have at least 10 regular members
+* Have received TAC approval of the readme.md per `Incubating` requirements above
+* Have met at least 4 times over a period of at least 2 months since becoming `Incubating`
+* Have at least 5 regular members from at least 3 different organizations attending regularly as
+recorded in meeting minutes.
 * Request TAC approval. TAC will vote to approve or provide constructive guidance
 
 ## Once `Active`:
--- a/working-group-lifecycle.md
+++ b/working-group-lifecycle.md
@@ -1,0 +1,67 @@
+# Working Group Life Cycle
+
+Contributors interested in starting a new working group are encouraged to first review the existing
+working groups and projects. It is very likely that an existing working group could be extended to
+cover the new topic. In that case please bring your time and energy into that existing working
+group. Your contributions are very welcome!
+
+However, if no existing effort covers the topic feel free to make use of the existing working group
+meetings, mail lists, and/or slack channels to solicit support for a new working group.
+
+Please note that running a working group successfully requires an extended effort.
+Be prepared to commit at least several hours per week for a year or more.
+
+The Working Group Life Cycle begins with establishing commitments from other interested
+contributors then making a proposal to the Technical Advisory Committee (TAC). If accepted, the
+Working Group will begin as `Incubating`. This is a boot strapping phase for the group to adopt
+OpenSSF community policies and norms. Once those are met the group becomes `Active` and will be
+fully recognized in OpenSSF. Either upon completing its objectives or no longer making forward
+progress a project will move to `End-of-Life`. The details for each phase follow.
+
+## To become `Incubating`
+
+* Proposal of scope for review by TAC
+    * This is to help ensure limited overlap with existing WG
+* Have met at least 5 times
+    * For these, meeting notes (or ideally recordings) are public
+* Have at least 5 interested individuals from at least 3 different organizations attending regularly
+* 1 TAC sponsor
+    * TAC sponsor agrees to attend WG meetings regularly
+    * TAC sponsor does not need to have a formal role in WG, e.g., chair
+    * TAC sponsor requests TAC approval 
+* TAC will vote to approve or provide constructive guidance
+
+## Once `Incubating`:
+
+* Operate as part of the OpenSSF adhering to community policies
+* Deliver quarterly reviews to TAC
+* Complete and request TAC approval of [readme.md](https://github.com/ossf/project-template/blob/main/README.md)
+* Meet on a regular cadence. Meetings are public, recorded, and on the calendar
+* Have access to community resources (Zooms, YouTube channels, GitHub, Slack channels, etc.)
+* Can request funding/other resources (subject to TAC/GB approval)
+    * NOTE: _At the time of this draft, funding and resources beyond collaboration tools have not been established in the OpenSSF. Working groups should expect that their main resource is the community contributions they are able to recruit._
+
+## To become `Active`:
+
+* Have met at least 20 times
+* Have at least 10 regular members
+* Request TAC approval. TAC will vote to approve or provide constructive guidance
+
+## Once `Active`
+
+* (same as incubating, plus)
+* Have at least annual goals and metrics for success
+
+## To become `End-of-Life`
+
+* Work has stopped
+    * The Working Group has completed its chartered deliverables
+    or
+    * The Working Group is no longer progressing on its deliverables as determined by the TAC
+* In either case the TAC will vote to formally end the Working Group
+
+## Once `End-of-Life`
+
+* All significant artifacts should be archived as appropriate
+* Meeting series removed from calendar, mail list closed, and any other administrative items
+  similarly put to closure as needed
--- a/working-group-lifecycle.md
+++ b/working-group-lifecycle.md
@@ -2,11 +2,11 @@
 
 Contributors interested in starting a new working group are encouraged to first review the existing
 working groups and projects. It is very likely that an existing working group could be extended to
-cover the new topic. In that case please bring your time and energy into that existing working
+cover the new topic. If this is the case, please bring your time and energy into that existing working
 group. Your contributions are very welcome!
 
 However, if no existing effort covers the topic feel free to make use of the existing working group
-meetings, mail lists, and/or slack channels to solicit support for a new working group.
+meetings, mailing lists, and/or Slack channels to solicit support for a new working group.
 
 Please note that running a working group successfully requires an extended effort.
 Be prepared to commit at least several hours per week for a year or more.
--- a/working-group-lifecycle.md
+++ b/working-group-lifecycle.md
@@ -18,7 +18,7 @@ OpenSSF community policies and norms. Once those are met the group becomes `Acti
 fully recognized in OpenSSF. Either upon completing its objectives or no longer making forward
 progress a project will move to `End-of-Life`. The details for each phase follow.
 
-## To become `Incubating`
+## To become `Incubating`:
 
 * Proposal of scope for review by TAC
     * This is to help ensure limited overlap with existing WG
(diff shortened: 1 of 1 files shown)


working-group-lifecycle.md
```diff
@@ -0,0 +1,67 @@
...
+* Deliver quarterly reviews to TAC
+* Complete and request TAC approval of [readme.md](https://github.com/ossf/project-template/blob/main/README.md)
+* Meet on a regular cadence. Meetings are public, recorded, and on the calendar
+* Have access to community resources (Zooms, YouTube channels, GitHub, Slack channels, etc.)
+* Can request funding/other resources (subject to TAC/GB approval)
+    * NOTE: _At the time of this draft, funding and resources beyond collaboration tools have not been established in the OpenSSF. Working groups should expect that their main resource is the community contributions they are able to recruit._
+
+## To become `Active`:
+
+* Have met at least 20 times
```
dlorenc, line 46: 20 times is basically an entire year - I'd be in favor of dropping this to 10.

working-group-lifecycle.md
```diff
@@ -0,0 +1,67 @@
...
+* Have at least 10 regular members
+* Request TAC approval. TAC will vote to approve or provide constructive guidance
+
+## Once `Active`
+
+* (same as incubating, plus)
+* Have at least annual goals and metrics for success
```
dlorenc, line 53: This one still doesn't make sense to me as a working group lead. What does a WG get in return for the extra overhead of producing annual goals and metrics?

I'm guessing this would get dropped over time and not strictly enforced, which is maybe not a bad thing.

dlorenc (commented):
Thanks Dan! I think this accurately captures where we left off yesterday.

working-group-lifecycle.md
jenniferfernick: I'm also in favour of it being a smaller number (such as 10). 

Could we modify this to be time or objective based, to reward progress? For example, we could try something like, "Have met at least 10 times and/or have published at least one major tool, paper, or library"

working-group-lifecycle.md
```diff
@@ -0,0 +1,67 @@
...
+The Working Group Life Cycle begins with establishing commitments from other interested
+contributors then making a proposal to the Technical Advisory Committee (TAC). If accepted, the
+Working Group will begin as `Incubating`. This is a boot strapping phase for the group to adopt
+OpenSSF community policies and norms. Once those are met the group becomes `Active` and will be
+fully recognized in OpenSSF. Either upon completing its objectives or no longer making forward
+progress a project will move to `End-of-Life`. The details for each phase follow.
+
+## To become `Incubating`
+
+* Proposal of scope for review by TAC
```
jenniferfernick, line 23: I would love if we could provide some "seed" resources to incubating projects and lay out a guide for how to access each one. This is arguably the time when they'll need our support the most. 

What subset of the following could we offer without too much admin overhead?  
"Have access to community resources (Zooms, YouTube channels, GitHub, Slack channels, etc.)"

estesp:
Minor point on formatting; 2 of the 6 headings for state changes end with a colon; 4 do not. I don't have a strong preference either way, but should be consistent.

Question: for the proposed numeric checks (5 people, 3 orgs; or 10 people involved, etc.) do we have any data on current WGs and where they sit against those criteria? Given I personally haven't been involved in the WGs to date, I don't have a sense for where that bar sits with respect to existing groups. Given that, it's hard to assess against my gut feel that these might be higher than expected. For example, there are some vibrant OSS projects with 5 or fewer people actively contributing.

Corollary: do we want to say anything about "judgement calls" for the TAC decision process and that the suggested items are effectively markers but not hard and fast rules?

dcmiddle:
See latest commit (and commit message) for reducing the meeting count and attendance numbers.
For rationale of separate active and incubating stages see the last paragraph of the intro, i.e., there's a checkpoint to ensure incubation completed correctly.
It's not a big deal either way but it's probably beneficial to have a checkpoint.

working-group-lifecycle.md
dlorenc: +1! What do you have in mind for these resources?

dlorenc:
Thanks for the updates Dan! I'm happy with this as a starting point, we can always iterate as we go.

One thing that got dropped somewhere between the initial discussion in #13 and now appears to be any kind of explanation on what it means to be a working group in either of the levels. From a practical perspective - what can a working group do, other than run meetings? This should cover stuff like:

- Can an active working group publish a white paper without tac approval? 
- What about an incubating one?
- Can a working group create repositories? When does it make sense to put these "in a WG" vs. treat them as new, standalone things?
- What about fundraising/soliciting donations?
- Where should repos/code "live"? This org or somewhere else?

I had some thoughts higher up in #13, but I think it's probably ok to punt on this for now. I do think we probably want to revisit that aspect before starting any reviews of the existing working groups though. For the time being the TAC has been erring on the side of minimal oversight/maximum autonomy which is great with me. It's hard to design governance in the abstract, so observing what the WGs are doing now will make this easier later. We're still small and it's still relatively easy to keep an eye on things.

Any thoughts from others?

JonZeolla:
As someone new to the OSSF and looking to get involved in some WGs I did find it helpful that there was a maintained index [under the tac](https://github.com/ossf/tac#technical-initiatives)  but it's possible not everyone will look at the tac README.  To keep discoverablility high I would suggest keeping all the repos under the OSSF org, maintain a working group topic or naming standard, and then maybe linking to a filtered query from the openssf.org website

working-group-lifecycle.md
david-a-wheeler: We should probably add one additional state, `proposed`, for a proposed WG that hasn't entered the `incubating` state.

dlorenc (approved):

estesp (approved):
LGTM

christo4ferris:
This is fine for a working group, I suppose, though I am still unclear about things such as who has commit privs for the created repository or how the WG intends to manage that over time. However, we now have a mix of projects (repos) and working groups, and no clear distinction between them. We also have no guidance as to what the minimum expectations are for the repository (things such as a README.md, CONTRIBUTING.md, LICENSE, MAINTAINERS.md, etc). I also don't understand how number of meetings and diversity of representation make a WG "active". The bar seems awfully low. How about ensuring that their meetings are advertised on the community calendar, that their agenda and minutes are posted along with recordings, etc. How about assessing whether the WG is actually producing something of value?

christo4ferris:
> Thanks for the updates Dan! I'm happy with this as a starting point, we can always iterate as we go.
> 
> One thing that got dropped somewhere between the initial discussion in #13 and now appears to be any kind of explanation on what it means to be a working group in either of the levels. From a practical perspective - what can a working group do, other than run meetings? This should cover stuff like:
> 
>     * Can an active working group publish a white paper without tac approval?
> 
>     * What about an incubating one?
> 
>     * Can a working group create repositories? When does it make sense to put these "in a WG" vs. treat them as new, standalone things?
> 
>     * What about fundraising/soliciting donations?
> 
>     * Where should repos/code "live"? This org or somewhere else?
> 
> 
> I had some thoughts higher up in #13, but I think it's probably ok to punt on this for now. I do think we probably want to revisit that aspect before starting any reviews of the existing working groups though. For the time being the TAC has been erring on the side of minimal oversight/maximum autonomy which is great with me. It's hard to design governance in the abstract, so observing what the WGs are doing now will make this easier later. We're still small and it's still relatively easy to keep an eye on things.
> 
> Any thoughts from others?

I really don't think it wise to punt on these items. They will only result in confusion. We have a number of incubating WGs now. Can they create repositories without oversight? How are they managed? have we policies for best practices for how a repository is structured and governed? Way too many open switches.

estesp:
@christo4ferris the disconnection between meeting minutes and GH issues/activity is at fault here. The TAC acknowledged all those items as needing further discussion, but getting a framework moving forward was deemed an important first step while we take the time to dig into more detail on the "how"/"what"/etc. of WG lifecycle activities. Given we are doing this as a GH markdown guide, iteration is relatively easy and PRs can continue to build on this initial framework.

dcmiddle:
See https://github.com/ossf/tac/issues/44 for what Active gets you.

rhaning (approved):

estesp (approved):
LGTM

working-group-lifecycle.md
```diff
@@ -0,0 +1,71 @@
+# Working Group Life Cycle
+
+Contributors interested in starting a new working group are encouraged to first review the existing
```
mayakacz, line 3:
```suggestion

## Starting a new working group

Are you interested in starting a new working group? Contributors interested in starting a new working group are encouraged to first review the existing
```

working-group-lifecycle.md
```diff
@@ -0,0 +1,71 @@
...
+working groups and projects. It is very likely that an existing working group could be extended to
+cover the new topic. In that case please bring your time and energy into that existing working
```
mayakacz, line 5:
```suggestion
cover the new topic. If this is the case, please bring your time and energy into that existing working
```

working-group-lifecycle.md
```diff
@@ -0,0 +1,71 @@
...
+group. Your contributions are very welcome!
+
+However, if no existing effort covers the topic feel free to make use of the existing working group
+meetings, mail lists, and/or slack channels to solicit support for a new working group.
```
mayakacz, line 9:
```suggestion
meetings, mailing lists, and/or Slack channels to solicit support for a new working group.
```

working-group-lifecycle.md
```diff
@@ -0,0 +1,71 @@
...
+
+Please note that running a working group successfully requires an extended effort.
+Be prepared to commit at least several hours per week for a year or more.
+
+The Working Group Life Cycle begins with establishing commitments from other interested
```
mayakacz, line 14: Feels like this should be at the top maybe?

working-group-lifecycle.md
mayakacz: > We should probably add one additional state, `proposed`, for a proposed WG that hasn't entered the `incubating` state.

I think I'm fine with just going straight to `incubating`.

> What subset of the following could we offer without too much admin overhead?
"Have access to community resources (Zooms, YouTube channels, GitHub, Slack channels, etc.)"

Now that that's included in the preamble, I don't think we need to include it here. Alternatively, something like "Can use community resources to organize, but do not have dedicated channels"

working-group-lifecycle.md
mayakacz: I think I originally suggested this because we wanted WGs to set goals and have a vision of what they're working on. But yeah, I agree with you that this is not necessarily the right thing to aim for - I also don't think there are necessarily expectations around the goals being met.
Maybe there's nothing 'additional' to do once you're active, as long as you keep doing the above. I'd be OK to remove.

dlorenc:
OK! This is approved now with 4. Ready to merge.

Issue #40, opened 2020-12-01
TAC to define 3-5 year technical vision


rhaning:
The TAC is to define the Technical Vision for OpenSSF for the next 3-5 years. An annual roadmap (#37) of detailed work will be created separately.

**Timeline**:

- Review Draft at TAC meeting December 15
- Review Final at TAC meeting January 12
- Press Release Final January 15 (to be confirmed)

**Guidelines/ideas for defining the vision**:

1. Aspirational and motivational.
2. Incorporate current work being done in OpenSSF and also broad enough to incorporate future, related works.
3. What has changed in 3-5 years because the OpenSSF exists?

**Working Group categorization**:

1. Educate – Best Practices
2. Inform - Identifying Security Threats, Vulnerability Disclosure, Digital Identity Attestation
3. Protect – Vulnerability Disclosure , Digital Identity Attestation, Security Tooling, Securing Critical Projects

david-a-wheeler:
This may be useful: [What Is a Vision Statement?](https://www.businessnewsdaily.com/3882-vision-statement.html).

david-a-wheeler:
Per the TAC meeting, it may be useful to look at the charter (adding more technical meat but still aspirational): https://github.com/ossf/foundation/blob/main/Review%20Copy%20Only%20-%20Not%20for%20Execution_OpenSSF%20Participation%20Agreement%20and%20Charter%20(rev.%202020%2011-10-2020).pdf

david-a-wheeler:
Per the 2020-11-31 TAC meeting, the plan is to put proposed text in this issue & discuss it here in this issue. Final version will go into a document file on GitHub (e.g., TAC README.md or its own vision.md file); the final version will not be a Google docs document.

kaywilliams:
I created rough draft vision as a starting point for discussion.  I am copy pasting below, but I also created a google doc [here](https://docs.google.com/document/d/1iw01-grJgGtFQeWTI-LN59vKrJ4sfG6Igvz9pYqrFak/edit#), in case people want to put comments in the doc or experiment with revisions.  (Sorry, I know we discussed keeping all iteration in the GitHub issue, but it felt unnatural to me.)

-------------
OpenSSF Technical Vision

We envision a future where participants in the Open Source Software ecosystem focus on delivering high quality products and services, with security handled naturally and automatically in the background. Aspects of this vision include the following:

- Security administrators (from project maintainers to government regulators) can specify security policy in a manner that can be easily automated.
[...]

PR #47, opened 2021-01-06
Create technical-vision.md

kaywilliams:
See related issue https://github.com/ossf/tac/issues/40

### DIFF
--- a/technical-vision.md
+++ b/technical-vision.md
@@ -1,0 +1,9 @@
+# OpenSSF Technical Vision
+
+We envision a future where participants in the open source software ecosystem focus on delivering high quality products and services, with security handled proactively, automatically, and as a matter of course.
+- Developers can easily learn secure development practices, and are proactively guided by their tools to apply those practices and automatically informed when action is needed to prevent, remediate, or mitigate security issues.
+- Developers, auditors and regulators can create and easily distribute security policies that are enforced through tooling and automation, providing continuous assurance of the results.
+- Developers and researchers can identify security issues (including unintentional vulnerabilities and malicious software) and have this information swiftly flow backwards through the supply chain to someone who can rapidly address the issue.
+- All stakeholders can provide information and notifications about product defects, mitigations, quality, and supportability and have this information rapidly flow forward across the ecosystem system to all users, and users can rapidly update their software or implement mitigations as appropriate.
+
+The Open Source Security Foundation (OpenSSF) provides tools, services, training, infrastructure, and resources to achieve this vision.
--- a/technical-vision.md
+++ b/technical-vision.md
@@ -1,9 +1,9 @@
 # OpenSSF Technical Vision
 
-We envision a future where participants in the open source software ecosystem focus on delivering high quality products and services, with security handled proactively, automatically, and as a matter of course.
+We envision a future where participants in the open source software ecosystem focus on delivering high quality products and services, with security handled proactively and as a matter of course.
 - Developers can easily learn secure development practices, and are proactively guided by their tools to apply those practices and automatically informed when action is needed to prevent, remediate, or mitigate security issues.
 - Developers, auditors and regulators can create and easily distribute security policies that are enforced through tooling and automation, providing continuous assurance of the results.
 - Developers and researchers can identify security issues (including unintentional vulnerabilities and malicious software) and have this information swiftly flow backwards through the supply chain to someone who can rapidly address the issue.
-- All stakeholders can provide information and notifications about product defects, mitigations, quality, and supportability and have this information rapidly flow forward across the ecosystem system to all users, and users can rapidly update their software or implement mitigations as appropriate.
+- Community members can provide information and notifications about product defects, mitigations, quality, and supportability and have this information rapidly flow forward across the ecosystem system to all users, and users can rapidly update their software or implement mitigations as appropriate.
 
 The Open Source Security Foundation (OpenSSF) provides tools, services, training, infrastructure, and resources to achieve this vision.


technical-vision.md
```diff
@@ -0,0 +1,9 @@
+# OpenSSF Technical Vision
+
+We envision a future where participants in the open source software ecosystem focus on delivering high quality products and services, with security handled proactively, automatically, and as a matter of course.
```
lukehinds, line 3: could we change products to projects, open source is not always a product, but always projects (even when someone creates a product out of them). Perhaps drop services (if this is meant in the 'professional services' capacity), if its 'to serve' then we just need to tweak the text. 

For the rest, I think the statement on 9 needs to come first or we reword, as otherwise the prior sentences sound like current facts. 

>  Our vision at The Open Source Security Foundation (OpenSSF) is to provide tools, services, training, infrastructure, and resources to help:

> - Developers learn secure development practices, and are proactively guided by their tools to apply those practices and automatically informed when action is needed to prevent, remediate, or mitigate security issues.
> - Developers, auditors and regulators to create and easily distribute security policies that are enforced through tooling and automation, providing continuous assurance of the results.

One last nit, there might be a better word them stakeholders, I much prefer "The community"

lukehinds (changes requested):
some comments..

technical-vision.md
```diff
@@ -0,0 +1,9 @@
...
+- Developers can easily learn secure development practices, and are proactively guided by their tools to apply those practices and automatically informed when action is needed to prevent, remediate, or mitigate security issues.
+- Developers, auditors and regulators can create and easily distribute security policies that are enforced through tooling and automation, providing continuous assurance of the results.
+- Developers and researchers can identify security issues (including unintentional vulnerabilities and malicious software) and have this information swiftly flow backwards through the supply chain to someone who can rapidly address the issue.
+- All stakeholders can provide information and notifications about product defects, mitigations, quality, and supportability and have this information rapidly flow forward across the ecosystem system to all users, and users can rapidly update their software or implement mitigations as appropriate.
```
lukehinds, line 7: Perhaps instead of stakeholders, "The community", some might have no stake, but just want to contibute for reasons for altruism.

lukehinds, line 7: Perhaps instead of stakeholders, "The community", some might have no stake, but just want to contibute for reasons for altruism.

technical-vision.md
```diff
@@ -0,0 +1,9 @@
...
+
+The Open Source Security Foundation (OpenSSF) provides tools, services, training, infrastructure, and resources to achieve this vision.
```
lukehinds, line 9: For the rest, I think the statement on 9 needs to come first before the bullert points or we reword, as otherwise the prior statements sound like current facts. 

>  Our vision at The Open Source Security Foundation (OpenSSF) is to provide tools, services, training, infrastructure, and resources to help:

> - Developers learn secure development practices, and are proactively guided by their tools to apply those practices and automatically informed when action is needed to prevent, remediate, or mitigate security issues.
> - Developers, auditors and regulators to create and easily distribute security policies that are enforced through tooling and automation, providing continuous assurance of the results.

technical-vision.md
scovetta: One comment here -- I don't think it's the intent, but the language might be taken to imply that "all" security work is done "automatically", or that "the only security work that is done will be driven from automated tools". I think the intent is (rightly) that "tools/technologies/platforms are used to detect, remediate, or obviate as much security risk as possible", or something to that extent.

technical-vision.md
kaywilliams: We want to keep the focus on the future we envision. The last sentence is a short nod to how OpenSSF specifically contributes to this future.  Another way to say this is that we are writing a vision statement rather than a mission statement.

technical-vision.md
kaywilliams: Good comment. Changing to 'All community members'...

technical-vision.md
kaywilliams: How about "We envision a future where participants in the open source software ecosystem focus on delivering high quality products and services, with security handled proactively and as a matter of course."  This reduces the focus on automation. I prefer not to narrow the scope to 'projects' since the broader ecosystem includes projects, packages, products, services, etc.  We could use a word like 'offerings' to encompass them all.  But for a general audience, I think the phrase 'products and services' is more natural wording.

lukehinds: I still think we need to remove products, as a lot of the eco-system are not products. They are software projects started as a labour of love , which have evolved into wide use (including the mother of them all, Linux). Perhaps I am getting hung up on semantics. 

I think this should solve it..

> "We envision a future where participants in the open source software ecosystem focus on delivering high quality secure code and services, with security handled proactively and as a matter of course."

technical-vision.md
lukehinds: ack, I don't mind the order too much. It's more that when I read it,  it sounds like the vision is the current status.

david-a-wheeler:
I’d use an Oxford comma and change:
> Developers, auditors and regulators
to:
> Developers, auditors, and regulators

david-a-wheeler:
One thing that’s not so obvious is that we need to get development tools & libraries modified so that they’re secure by default (the “obvious/default thing to do” is also the secure thing to do). I’d like to think that’s obliquely covered by the phrase "with security handled proactively and as a matter of course”.

But maybe that’s not clear enough. We could emphasize that by adding a bullet after the first one, something like this:

2. Developers create tools and libraries that, where practicable, are secure by default.

technical-vision.md
rhaning: I agree with Luke on this. In the context of open source, either 'project' and/or 'code' is more appropriate as not all OSS will be a product or a service. It can be a tool, a component or library. A project/code is all encompassing, 'product' is narrow IMO.

kaywilliams: How about this wording? 

> "We envision a future where participants in the open source ecosystem use and share high quality software, with security handled proactively and as a matter of course:"

Here we use the general word 'software'.

technical-vision.md
kaywilliams: How about this?  I added a colon to the intro sentence to clarify that the subsequent bullets are part of the future we envision, rather than standalone items.

>'We envision a future where participants in the open source ecosystem use and share high quality software, with security handled proactively and as a matter of course:'

Between that and the title of the document as "OpenSSF Technical Vision" this should be clear? Open to other suggestions.

lukehinds: LGTM!

lukehinds (approved):
+1 from me.

dlorenc (approved):

david-a-wheeler:
Looks good!

mayakacz (approved):

rhaning (approved):

estesp (approved):
LGTM

PR #48, opened 2021-01-12
Update issue template for new org/team

mayakacz:
Draft for an issue template for adding org/team members.
* A user can request to be added to the GitHub org
* An org can request that we create a new team for their org within the GitHub org

I think the use case we still need to figure out is new individuals in an existing member org. As far as I can tell, unless we make those who are team maintainers for the teams also org admins, they can't add *individuals* to the GitHub org.

### DIFF
--- a/.github/ISSUE_TEMPLATE/join-the-openssf-github-org.md
+++ b/.github/ISSUE_TEMPLATE/join-the-openssf-github-org.md
@@ -1,0 +1,28 @@
+---
+name: Join the OpenSSF GitHub org
+about: For members requesting access to the OpenSSF GitHub org
+title: "[GitHub org request] Add organization/team"
+labels: ''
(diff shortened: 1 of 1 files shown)


rhaning (approved):

mayakacz:
@rhaning Are you OK with someone from a member org requesting individual access, and then getting added to a team?

rhaning:
> @rhaning Are you OK with someone from a member org requesting individual access, and then getting added to a team?

@mayakacz In this context I think it's okay. Participating in projects doesn't require organizational or OpenSSF membership, so it seems reasonable to allow this. Is there a potential issue that you're concerned about?

mayakacz:
> Is there a potential issue that you're concerned about?

No, just that it's burdensome.

PR #75, opened 2022-01-19
Add SCIR and TAC 2022 nominations

david-a-wheeler:
Add PDF files with SCIR and TAC nominations.
This was originally in Google Sheets; I tried converting
them to HTML and markdown, but the first results were ugly
and I don't have a lot of time to do the conversion.
For this purpose I think a PDF is okay.

Signed-off-by: David A. Wheeler <dwheeler@dwheeler.com>


david-a-wheeler:
I'm adding this. We want to make the voting process maximally transparent & record this info. Note that the candidate statements will also be directly presented in the voting process, this isn't the only place people can see them.

PR #96, opened 2022-04-12
Separate WGs and Projects lifecycles Take 2

lehors:
This is a different take on the separation of WGs and Projects which tries to stick to the reality OpenSSF finds itself in and where we have many initiatives called "projects" that are not software focused.

This approach leverages the existing WG lifecycle and leverages the new lifecycle
inherited from CNCF for software focused projects while introducing a new
yet-to-be-defined lifecycle for non-software focused projects.

Signed-off-by: Arnaud J Le Hors <lehors@us.ibm.com>

IMPORTANT NOTE: This PR is an alternative to PR #95. Only one of them should be merged.

Additional Note: If adopted, this approach should be followed by an update of the OpenSSF charter to match this model.

PR #95, opened 2022-04-11
Separate WGs and Projects lifecycles

lehors:
This leverages the existing WG lifecycle and segregates the new lifecycle
inherited from CNCF to projects which are directly reporting to the TAC
or a parent WG.

Signed-off-by: Arnaud J Le Hors <lehors@us.ibm.com>

PR #98, opened 2022-04-21
Swap WG and SIG, and reserve Projects for software

lehors:
This PR incorporates the new proposed terminology and structure with the change of WGs into SIGs, and Projects being reserved for software focused efforts to better align with other LF efforts.

Various important details regarding the governance and decision making still need to be figured out.

This is an evolution off of PR 95 and needs to be considered as an alternative to both PR #95 and #96.


process/working-group-lifecycle.md
```diff
@@ -0,0 +1,13 @@
+# Working Group (WG) Life Cycle
+
+Working Groups are under the direct governance of a Special Interest Group (SIG) which can create and close them as necessary and appropriate. WGs are expected to have a very specific goal and not necessarily live for a very long time.
```
dlorenc, line 3: Same as above. I'd rather not include this file here.

lehors:
Given that PR#95 has now been merged and nobody seems to really have the appetite for swapping the names due to the high cost this would incur I'm closing this.

PR #102, opened 2022-05-10
Adding a minimal lifecycle for SIGs

lehors:
This PR fills in a gap left by PR#95. I recommend merging this to add the file and then argue over the exact amount of control and oversight there should actually be and by whom.

Signed-off-by: Arnaud J Le Hors <lehors@us.ibm.com>


AevaOnline:
LGTM

PR #109, opened 2022-06-15
Clarify that Sandbox is optional

lehors:
This PR removes the ambiguity that currently exists in the document by explicitly allowing projects to enter incubation directly.

Signed-off-by: Arnaud J Le Hors <lehors@us.ibm.com>


annabellegoth2boss:
I would suggest documenting it as, "...may enter Incubating or Graduating" to give the option for a mature project that lives elsewhere or independently to join the OSSF.

dlorenc:
My 2c: I agree with Anne, no need to force projects through a specific level.

AevaOnline:
I'd prefer to avoid over-optimizing before we have established precedent and well-oiled process for evaluating the different levels. If we allow projects to apply directly to Graduated, we could end up roadblocked on the technical and community-maturity evaluation of a project that we don't have any engagement in, and then mired in unknowns.

Better to walk before we run...

So, this patch LGTM as is.

PR #114, opened 2022-07-11
Fixing link to OpenSSF Charter

gkunz:
The link to the OpenSSF technical charter in the README is broken. This
changes the link to point to the new markdown version of the charter in
the foundation repository.

Signed-off-by: Georg Kunz <georg.kunz@ericsson.com>

PR #116, opened 2022-07-13
fix link to charter PDF

bobcallaway:
The charter PDF link was previously broken - fixing it to a working link.

Signed-off-by: Bob Callaway <bcallaway@google.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Meetings are also recorded and posted to the [OpenSSF YouTube channel](https://w
 
 ## Charter
 
-The TAC is chartered as part of the [Open Source Security Foundation Charter](https://github.com/ossf/foundation/blob/main/Review%20Copy%20Only%20-%20Not%20for%20Execution_OpenSSF%20Participation%20Agreement%20and%20Charter%20(rev.%202020%2009%2011).pdf).
+The TAC is chartered as part of the [Open Source Security Foundation Charter](https://cdn.platform.linuxfoundation.org/agreements/openssf.pdf).
 
 ## Technical Initiatives
 


inferno-chromium (approved):
@david-a-wheeler @brianbehlendorf - this one says "review copy not for execution", David, Brian - is there any other link ?

david-a-wheeler:
@inferno-chromium - that's the only link I know of.

The "review copy not for execution" text is intentional. My understanding is that we don't want organizations signing that & sending it via email. If an organization wants to join, they need to go through a specific automated process that records & verifies various things.

inferno-chromium:
> @inferno-chromium - that's the only link I know of.
> 
> The "review copy not for execution" text is intentional. My understanding is that we don't want organizations signing that & sending it via email. If an organization wants to join, they need to go through a specific automated process that records & verifies various things.

SGTM, thanks!

lehors:
As I pointed out elsewhere, the problem is that what's referred to as the charter is not actually the charter. It's a sample membership agreement which includes the charter as an annex. We ought to make the charter available on its own.

README.md
```diff
@@ -35,7 +35,7 @@ Meetings are also recorded and posted to the [OpenSSF YouTube channel](https://w
 
 ## Charter
 
-The TAC is chartered as part of the [Open Source Security Foundation Charter](https://github.com/ossf/foundation/blob/main/Review%20Copy%20Only%20-%20Not%20for%20Execution_OpenSSF%20Participation%20Agreement%20and%20Charter%20(rev.%202020%2009%2011).pdf).
+The TAC is chartered as part of the [Open Source Security Foundation Charter](https://cdn.platform.linuxfoundation.org/agreements/openssf.pdf).
```
lehors, line 38: The canonical address is actually: https://charter.openssf.org/

lehors (changes requested):

brianbehlendorf:
On Fri, 15 Jul 2022, Arnaud J Le Hors wrote:
> As I pointed out elsewhere, the problem is that what's referred to as 
> the charter is not actually the charter. It's a sample membership 
> agreement which includes the charter as an annex. We ought to make the 
> charter available on its own.

The charter in the membership participation agreement is not a mere annex 
or exhibit for informational purposes, it's a part of the contractual 
commitment between parties who sign onto the agreement and the LF in how 
we manage the project. If the TAC wishes to maintain a copy of this 
charter in their repo, so that they don't need to see the "For Review 
Only" watermark or the other membership-specific content, fine, but it's 
akin to keeping a copy of the U.S. Constitution in the repo as well.

lehors:
> On Fri, 15 Jul 2022, Arnaud J Le Hors wrote: As I pointed out elsewhere, the problem is that what's referred to as the charter is not actually the charter. It's a sample membership agreement which includes the charter as an annex. We ought to make the charter available on its own.
> The charter in the membership participation agreement is not a mere annex or exhibit for informational purposes, it's a part of the contractual commitment between parties who sign onto the agreement and the LF in how we manage the project. If the TAC wishes to maintain a copy of this charter in their repo, so that they don't need to see the "For Review Only" watermark or the other membership-specific content, fine, but it's akin to keeping a copy of the U.S. Constitution in the repo as well.

I did say that the charter in annex is part of the membership agreement. But what's the status of the following, just to point to a few examples from a quick search ?

https://www.hyperledger.org/about/charter-2
https://www.lfedge.org/wp-content/uploads/2021/05/LF-Edge-Charter-rev.-2021-05-25.pdf
https://github.com/cncf/foundation/blob/main/charter.md

Are these too mere copies of the constitution ?
I think we can at least agree that this is one of those things LF doesn't handle consistently. (I guess you could argue that it's consistent with the moto: "There is no such things as the LF way!" ;-)

gkunz:
Hi all, @AevaOnline actually has a PR open (https://github.com/ossf/foundation/pull/25), which adds a markdown version of the charter to the foundation repository. I think this solves some of the problems outlined here.

I have created a draft PR (https://github.com/ossf/tac/pull/114) which is supposed to change the link to the markdown version of the charter once @AevaOnline's PR lands.

Fixing the link in the meantime is of course useful, but ideally we should change it to the MD version eventually.

lehors:
This should be updated to simply point to the charter which is now available in plain text at: https://openssf.org/about/charter/ or closed if @gkunz gets to udpate PR #114 first.

inferno-chromium:
Closing this since https://github.com/ossf/tac/pull/114 has landed.

PR #141, opened 2023-02-28
clarify terms of appointed and elected members

tpepper:
Given there's been discussion of whether the charter perhaps evolves to stagger terms and the election/appointing cadence is still early in its operation, it is useful to be able to see which members came into the board at what time.  Perhaps eventually this shows via the git history, but that too takes time evolve and settle and is less transparent for a typical reader.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,16 +22,17 @@ Meetings are also recorded and posted to the [OpenSSF YouTube channel](https://w
 
 ## Members
 
-| Name          | Email                          | Organization |
-| ------------- | ------------------------------ | ------------ |
-| Abhishek Arya | aarya@google.com               | Google       |
-| Aeva Black    | aeva.black@microsoft.com       | Microsoft    |
-| Bob Callaway  | bcallaway@google.com           | Google       |
-| CRob Robinson | christopher.robinson@intel.com | Intel        |
-| Dan Lorenc    | dlorenc@chainguard.dev         | Chainguard   |
-| Josh Bressers | josh.bressers@anchore.com      | Anchore      |
-| Luke Hinds    | lhinds@redhat.com              | Red Hat      |
+| Name          | Email                          | Organization | Term                     |
+| ------------- | ------------------------------ | ------------ | -------------------------|
+| Abhishek Arya | aarya@google.com               | Google       | April 2022 - March 2023  |
+| Aeva Black    | aeva.black@microsoft.com       | Microsoft    | April 2022 - March 2023\* |
+| Bob Callaway  | bcallaway@google.com           | Google       | April 2022 - March 2023  |
+| CRob Robinson | christopher.robinson@intel.com | Intel        | April 2022 - March 2023\* |
+| Dan Lorenc    | dlorenc@chainguard.dev         | Chainguard   | April 2022 - March 2023  |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,15 +22,17 @@ Meetings are also recorded and posted to the [OpenSSF YouTube channel](https://w
...
-| Name          | Email                          | Organization |
-| ------------- | ------------------------------ | ------------ |
-| Abhishek Arya | aarya@google.com               | Google       |
-| Aeva Black    | aeva.black@microsoft.com       | Microsoft    |
-| Bob Callaway  | bcallaway@google.com           | Google       |
-| CRob Robinson | christopher.robinson@intel.com | Intel        |
-| Dan Lorenc    | dlorenc@chainguard.dev         | Chainguard   |
-| Josh Bressers | josh.bressers@anchore.com      | Anchore      |
-| Luke Hinds    | lhinds@redhat.com              | Red Hat      |
+| Name          | Email                          | Organization | Term     |
```
joshuagl, line 25:
```suggestion
| Name          | Email                          | Organization | Term     |
```
Should we say "Term end" here?

joshuagl (commented):
This is both informative and improves transparency. 👍

README.md
tpepper: I was trying to decide if it should be term start-end year like 2022-2023 or put something like "April 2022 - March 2023".  And then the line was wrapping in my text editor and I just went terse :)

README.md
```diff
@@ -22,15 +22,17 @@ Meetings are also recorded and posted to the [OpenSSF YouTube channel](https://w
...
+| ------------- | ------------------------------ | ------------ | ---------|
+| Abhishek Arya | aarya@google.com               | Google       | 2022 (E) |
+| Aeva Black    | aeva.black@microsoft.com       | Microsoft    | 2022 (A) |
+| Bob Callaway  | bcallaway@google.com           | Google       | 2022 (E) |
+| CRob Robinson | christopher.robinson@intel.com | Intel        | 2022 (A) |
+| Dan Lorenc    | dlorenc@chainguard.dev         | Chainguard   | 2022 (E) |
+| Josh Bressers | josh.bressers@anchore.com      | Anchore      | 2022 (A) |
+| Luke Hinds    | lhinds@redhat.com              | Red Hat      | 2022 (E) |
+
+NOTE: In this table's Term column "(A)" and "(E)" respectively denote Governing Board appointed and community elected members, with election results and appointments announce annually in the month of April.
```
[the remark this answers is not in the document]
tpepper: Amended.

README.md
joshuagl: I would favour "April 2022 - March 2023". It feels like more specific is better for every use-case I can imagine.

joshuagl (approved):

lehors (approved):
LGTM

inferno-chromium:
I agree with @joshuagl comment, lets add the specific term period time.

Also, i think instead of A, E, i would prefer GB, and CE(or just C)

tpepper:
Updated based on @joshuagl and @inferno-chromium comments.  Still fits a reasonable width page and is definitely more readable.  What do folks think of this iteration?

bobcallaway (approved):
lgtm

joshuagl (approved):
Looks good, thanks

inferno-chromium (approved):

Issue #88, opened 2022-03-17
TAC repo uses the wrong license (Apache vs CC)


lehors:
The LICENSE file in the TAC repo currently contains the Apache Software license. Given that the TAC isn't going to develop code though it should contain the Creative Common Attribution license instead. This will be applicable to any documentation the TAC develops.

joshbressers:
I approve of this change mechanically, but to avoid future legal concerns, do we need to get a sign off or approval from someone with legal authority at the Linux Foundation?

I suspect all contributions up to this point are owned by the contributors and covered under Apache 2. I'm unsure how that would affect this move.

@david-a-wheeler @brianbehlendorf perhaps one of you can weigh in

david-a-wheeler:
@joshbressers - fair concern.

I'm not a lawyer. However, there isn't much text under the TAC currently, and the individual contributions are likely de minimus. We could probably just switch. However, if we want to be cautious, we could ask all the current & previous TAC members to approve release under CC-BY-4.0 via email & then we're fully covered for their contributions. I don't think anyone else has contributed directly to TAC stuff.

To avoid future problems, perhaps the LICENSE text should say:

> Unless otherwise noted, all text released directly from the OpenSSF TAC is licensed under the Creative Commons Attribution license (CC-BY-4.0), source code is released under the Apache-2.0 or MIT license, and standards are released under Community Specification License. For the license of a specific OpenSSF project, see that project. See the OpenSSF Charter for OpenSSF license policies.

Then under the file LICENSES, add each license using its SPDX id as its filename. That's the usual way to do this. Community Specification license Version 1.0 is available at https://github.com/CommunitySpecification/1.0 

BTW, the SPDX id for the current Creative Commons Attribution license is CC-BY-4.0 <https://spdx.org/licenses/>.

lehors:
I agree that @joshbressers raises a valid question. We don't need approval to switch license from all tac members though, only those who have contributed content to this repo. I suspect this is a lot fewer. And the answer is in the git log. :-)
I'm away from my computer right now but could look into it.
@david-a-wheeler As for the actual text, I don't think we can have a "this license or that license" kind of statement. It's got to be definitive otherwise one wouldn't know which applies to any content without an explicit embedded license (or SPDX I'd).

david-a-wheeler:
@lehors - no law requires machine-processable text in a LICENSE file. My worry is that if we drop CC-BY into LICENSE, we'll have the same problem if the TAC generates something else.

[...]

PR #143, opened 2023-03-10
Fix license to match charter

lehors:
This is a new attempt to address issue #88 which doesn't require chasing former contributors to get their agreement to change the license. This keeps the Apache license for prior work and only change the license for new one.
This replaces PR #88.


LICENSE
```diff
@@ -1,3 +1,16 @@
+Atrifacts and contributions to this repository prior to March 10, 2023
```
inferno-chromium, line 1: typo:Artifacts

inferno-chromium (approved):

LICENSE
lehors: Fixed. Thanks.

Issue #161, opened 2023-05-13
Update SIG processes and ensure consistency & alignment with WG & project processes


SecurityCRob:
TAC has good documentation around new project creation & lifecycle & working group lifecycle- https://github.com/ossf/tac/blob/main/process/project-lifecycle.md , https://github.com/ossf/tac/blob/main/process/working-group-lifecycle.md

Comparability, SIG documentation is lacking -  https://github.com/ossf/tac/blob/main/process/sig-lifecycle.md

No documentation exists today for Special Interest Funds (SIFs) or Affiliated Projects (APs)

The TAC should ensure clearly documented processes exist that describe:
- Special interest Groups(SIG) have an adoption and lifecycle process that describes stages from beginning to possible eventual retirement of these groups
- gives/gets for each level of the SIG lifecycle should be 
- Document who evaluates and votes upon creation/retirement/modification of these groups - TAC, WG, other?
- SIFs and APs should be clearly documented for proposal, adoption, and lifecycle processes
- documentation/language updated in WG/SIG charters, readme's, & security.md's that accurately describes desired processes for membership criteria, membership ladders/rewarding contributions, documentation how group votes are conducted and eligibility requirements

SecurityCRob:
The guidance does not need to be heavy, but should inform folks any required steps or things to care about so they can run with it and go do great things.

steiza:
We might need to update https://github.com/ossf/tac/blob/main/organizational-structure-overview.md as part of this work.

Currently that document describes SIGs as reporting to a WG, not the TAC (although even if we keep that structure, we can of course provide lightweight guidance on adoption criteria and lifecycle process).

AevaOnline:
The existing documentation does not cover SIFs or APs, as my previous attempts to quantify those were post-hoc and encountered organizational challenges.

On the topic of WGs and SIGs, refining the process from last year, and normalizing it across WGs and SIGs, seems entirely appropriate to me. Are there specific areas of friction or fragmentation that folks would like to address?

hythloda:
This was brought up in the TAC with @torgo @ware @bobcallaway @steiza and @hythloda 
Do want a zoom meeting to get this started? Or async?

ware:
I'll defer to others.  I can do either.

lehors:
> We might need to update https://github.com/ossf/tac/blob/main/organizational-structure-overview.md as part of this work.
> 
[...]

Issue #162, opened 2023-05-13
Foundation groups documentation audit 


SecurityCRob:
A short-lived working committee should be created under the TAC for the purpose of conducting a review to ensure existence, consistency, and accuracy for all Foundation group documentations (TAC, WG, SIG, SIF, AP, Committees, etc.).  This should include, but not be limited to the following artifacts:
- clear readme.md file that provides overview of group, with meeting times, communication channels, all active & past work, and areas where contributions are desired, any sub-groups affiliated with the higher-level working group, group leader(s), designated TAC liaison that assists the group 
- a clearly discoverable list of active members, their project level (maintainer, collaborator, contributor, etc) as well as membership criteria and voting procedures
- an up-to-date and approved group charter.md
- a security.md file that documents project defect and vulnerability reporting process (sourced from approved foundation template)
- other artifacts or documentation that is deemed necessary by the committee

hythloda:
This was brought up in the TAC with @steiza @jchestershopify @lehors 
Do want a zoom meeting to get this started? Or async?

jchester:
A nit: the best handle to use for me right now is this one; @jchestershopify is (as the name suggests) tied to my previous work for Shopify.

di:
For projects of the OpenSSF, should we add a "contributor ladder" to the list of artifacts? AFAIK [Allstar is the only project that approaches this](https://github.com/ossf/allstar/blob/main/contributor-ladder.md), I think this would be good to roughly standardize across all projects.

di:
Aha, thanks. (We might also want to align on what this document is called, to help with discoverability 😉)

di:
Perhaps this is also an opportunity to make a [template repository](https://docs.github.com/en/repositories/creating-and-managing-repositories/creating-a-template-repository) that can be used for new GitHub repositories?

hythloda:
We have a [WIP proposal](https://docs.google.com/spreadsheets/d/140czMjqhe9YxLBTx4rWkjHxWEWqD_tVQ1s8l9Q2PE2g/edit#gid=0) for what sections each should have. We would love comments!

jeffmendoza:
Allstar can enforce security.md and branch protection currently. We can add more features as needed as well.

SecurityCRob:
As TIs are reporting in quarterly, part of that process is a docs review/check to ensure that group has completed all the necessary tasks.  This review should be complete by the end of Q3.

PR #182, opened 2023-07-07
Update SIG lifecycle doc

steiza:
For https://github.com/ossf/tac/issues/161

Based on questions raised and community discussion

### DIFF
--- a/process/sig-lifecycle.md
+++ b/process/sig-lifecycle.md
@@ -6,6 +6,8 @@ It is expected that the primary output of a SIG is not software. If the primary 
 
 SIG process should be minimal, however we do need some process to at least ensure we have an accurate list of all the active SIGs in the OpenSSF. This document uses "must" to describe what items are required, "should" to suggest items that should be strongly considered (but not required), and "may" for suggested guidance.
 
+The SIG life cycle begins with interested contributors deciding to undertake this process, at which time the SIG is `Tentative`. A SIG becomes `Active` when a governing body agrees to take it on, at which point the SIG goes about achieving its goals and deliverables. Once that work is completed, or if the effort stalls out, the SIG becomes `End-of-Life` and some administrative cleanup will be done. The details for each phase follow.
+
 ## To become `Tentative`:
 
 * This is the default state of a new SIG that is not yet active
--- a/process/sig-lifecycle.md
+++ b/process/sig-lifecycle.md
@@ -20,7 +20,6 @@ The SIG life cycle begins with interested contributors deciding to undertake thi
 ## Once `Active`:
 
 * The governing body must list information about the SIG on its README, including current state, chairs, and a statement of focus, intent, goals, and/or deliverables
-  * SIGs may have a repository (prefixed with `sig-`) that include the information listed above
 * SIGs may have regular meetings separate from their governing body, if so:
   * They should appear on the OpenSSF calendar
   * They should have a document with upcoming agendas and notes from past meetings
--- a/process/sig-lifecycle.md
+++ b/process/sig-lifecycle.md
@@ -15,9 +15,9 @@ SIG process should be minimal with as much as possible delegated to the WG. Howe
 * A SIG must have a repository (prefixed with `sig-`) that includes the current state of the SIG, lists the chairs of the SIG, and a statement of the focus, intent, goals, and/or deliverables of the group
 * The chairs of a WG must agree to govern the SIG
   * The chairs of the WG should have its membership vote, with notice given at the meeting prior to the vote
-* The governing WG must list the active SIGs it governs, along with chairs and links to the repository of those SGs
+* The governing WG must list the active SIGs it governs, along with chairs and links to the repository of those SIGs
   * The governing WG should list previous SIGs that are now inactive
-* SIGs should have regular meetings
+* SIGs may have regular meetings
   * If these meetings are separate from their governing WG:
     * They should appear on the OpenSSF calendar
     * They should have a document with upcoming agendas and notes from past meetings
--- a/process/sig-lifecycle.md
+++ b/process/sig-lifecycle.md
@@ -1,10 +1,10 @@
 # Special Interest Group (SIG) Life Cycle
 
-Special Interest Groups are under the direct governance of their reporting Working Group (WG) and are bound to achieving a very specific goal. These groups may be terminated upon completion of their designating tasking, continued for larger and ongoing efforts, or otherwise subject to the governance, structure, and termination policies of the WG they are under. The creation of a SIG must dictate the focus, intent, goals, and deliverables as appropriate.
+Special Interest Groups are bound to achieving a very specific goal. These groups may be terminated upon completion of their designating tasking or continued for larger and ongoing efforts. The creation of a SIG must dictate the focus, intent, goals, and deliverable(s) as appropriate. SIGs may report to a WG or directly to the TAC as their governing body.
 
 It is expected that the primary output of a SIG is not software. If the primary output is software, the work should be organized as a [project](./project-lifecycle.md).
 
-SIG process should be minimal with as much as possible delegated to the WG. However we do need minimal process to at least ensure we have an accurate list of all the active SIGs in the OpenSSF. This document uses "must" to describe what items are required and "should" to suggest items that should be strongly considered (but not required).
+SIG process should be minimal, however we do need some process to at least ensure we have an accurate list of all the active SIGs in the OpenSSF. This document uses "must" to describe what items are required, "should" to suggest items that should be strongly considered (but not required), and "may" for suggested guidance.
 
 ## To become `Tentative`:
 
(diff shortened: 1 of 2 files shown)


process/sig-lifecycle.md
```diff
@@ -1,13 +1,30 @@
 # Special Interest Group (SIG) Life Cycle
 
-Special Interest Groups are under the direct governance of a Working Group (WG) which can create and close them as necessary and appropriate. SIGs are expected to have a very specific goals and objectives at the time of their creation which may or may not define if they are terminable and is entirely dependent on the nature of the SIG.
+Special Interest Groups are under the direct governance of their reporting Working Group (WG) and are bound to achieving a very specific goal. These groups may be terminated upon completion of their designating tasking, continued for larger and ongoing efforts, or otherwise subject to the governance, structure, and termination policies of the WG they are under. The creation of a SIG must dictate the focus, intent, goals, and deliverables as appropriate.
```
steiza, line 3: This language came from https://github.com/ossf/tac/blob/main/organizational-structure-overview.md

process/sig-lifecycle.md
```diff
@@ -1,13 +1,30 @@
...
 
-The lifecycle of SIGs is therefore minimal and as follows.
+It is expected that the primary output of a SIG is not software. If the primary output is software, the work should be organized as a [project](./project-lifecycle.md).
 
-## Active
+SIG process should be minimal with as much as possible delegated to the WG. However we do need minimal process to at least ensure we have an accurate list of all the active SIGs in the OpenSSF. This document uses "must" to describe what items are required and "should" to suggest items that should be strongly considered (but not required).
 
-* The SIG is actively pursuing its goal.
+## To become `Tentative`:
```
steiza, line 9: I landed on `Tentative` as a term that includes both the connotation that it's in an early phase of development, without trying to imply that it has endorsement (like `Proposed` might).

process/sig-lifecycle.md
```diff
@@ -1,13 +1,30 @@
...
 
-## Inactive
+* This is the default state of a new SIG that is not yet active
 
-* The SIG is no longer active, either it has completed its mission or the mission was abandoned. At which state, the WG the SIG is under will determine a termination status and record it within the corresponding repository.
+## To become `Active`:
+
+* A SIG must have a repository (prefixed with `sig-`) that includes the current state of the SIG, lists the chairs of the SIG, and a statement of the focus, intent, goals, and/or deliverables of the group
```
steiza, line 15: SLSA Positioning, Specification, and Tooling, as well as Vulnerability Autofix SIG do not have repositories today. If we like this item, we can go back and work with those SIGs to set that up.

process/sig-lifecycle.md
steiza: https://github.com/ossf/Diagrammers-Society today is a SIG that reports to the TAC. An explicit goal of this iteration is not to undo past decisions, but this would prevent new SIGs from forming under the TAC.

lehors: I think we should allow SIGs to report directly to the TAC and update everything accordingly.
There is no harm in doing this and I don't know that the restriction to have SIGs only report to a WG was intentional in the first place. I think it was simply what we had at the time.

process/sig-lifecycle.md
lehors: How about "Forming"?

process/sig-lifecycle.md
lehors: I don't think we should make it a requirement to have a repo for every SIG. I think it is perfectly fine for a SIG to simply have a page as part of their parent WG's repo.

process/sig-lifecycle.md
```diff
@@ -1,13 +1,30 @@
...
+* The chairs of a WG must agree to govern the SIG
+  * The chairs of the WG should have its membership vote, with notice given at the meeting prior to the vote
+* The governing WG must list the active SIGs it governs, along with chairs and links to the repository of those SGs
+  * The governing WG should list previous SIGs that are now inactive
+* SIGs should have regular meetings
```
lehors, line 20: Do we really want to require that?? This would mean a SIG can't just work asynchronously over slack and github for instance. I don't think that's right.
We do need to require some sustained activity so that we don't end up with zombie SIGs lying around but I don't think having meetings should be the right metric for this.

lehors (changes requested):
I must admit not to understand part of this PR. It enumerates how to transition to various states with defining them. I think we should have both, as in:

Tentative
....
To become Tentative
....
Active
....
To become Active
....
Inactive
....
To become Inactive
....

process/sig-lifecycle.md
steiza: > Each SIG must have a documented landing page, where one can learn more about the SIG, how to get involved, etc. but that should only be a dedicated repo if the repo will be otherwise active beyond providing that landing page.

I am definitely open to this direction, but help me understand - what would that landing page be? Would it be part of the WG's repository's README? Or a separate page in the WG's repository? Or something else?

process/sig-lifecycle.md
steiza: The "should" means that this is a recommendation and not a requirement - maybe the "should" should be more clear? Otherwise I'm happy to strike this item altogether if we don't think it should be a recommendation.

process/sig-lifecycle.md
steiza: Do note from https://github.com/ossf/tac/issues/162 we have a tentative list (@hythloda has [explicitly asked for feedback](https://github.com/ossf/tac/issues/162#issuecomment-1572589533)) on what we expect of Projects / SIGs today: https://docs.google.com/spreadsheets/d/140czMjqhe9YxLBTx4rWkjHxWEWqD_tVQ1s8l9Q2PE2g/edit#gid=821838029

di: > Would it be part of the WG's repository's README? Or a separate page in the WG's repository?

Either of these make sense to me. I think the important thing is that since a SIG must have a 1:1 relationship with a WG (or the TAC), the repo for that WG can list SIGs it is responsible for, with this information.

process/sig-lifecycle.md
di: I don't think this is strictly necessary for a SIG to be successful.

di, line 20:
```suggestion
* SIGs may have regular meetings
```

process/sig-lifecycle.md
steiza: I think `Forming` still could imply some level of endorsement - I'm going to stick with `Tentative`.

process/sig-lifecycle.md
steiza: I have updated the text!

process/sig-lifecycle.md
steiza: I have updated the text so SIG repositories are optional, and the information is listed on the governing body's (WG or TAC) README.

steiza:
> what we plan to do with existing SIGs that are not fully aligned with these requirements, i.e. haven't been explicitly agreed upon by a governing body

Good question! I had to go back and review my notes on [OpenSSF Workstream Taxonomy - Existing hierarchy](https://docs.google.com/document/d/1uONXSNOcyrT8IJsC3R52KiwE78IJAsddO9EZwURXazg/edit#heading=h.82vks3tar8h7).

I think all the SIGs already correspond to a WG, except for https://github.com/ossf/diagrammers-society/ which reports to the TAC (which we are updating the docs to allow).

The docs also says:
> This workstream will not undo previously made decisions

So my recommendation would be that we just update the TAC README to reflect this legacy decision.

di:
The example I was thinking of is the "Great Artifact Repository Audit" effort which is not listed in your taxonomy doc. Specifically, the Securing Software Repos WG hasn't voted to create the SIG (in fact the issue about it is still open: https://github.com/ossf/wg-securing-software-repos/issues/14), but AFAICT that work is a) ongoing regardless b) embodies a "SIG", at least in my opinion, and c) exists or should exist under that WG.

lehors (approved):

lehors (approved):
Thanks. The additional paragraph fulfills my request!

steiza:
> The example I was thinking of is the "Great Artifact Repository Audit" effort which is not listed in your taxonomy doc. Specifically, the Securing Software Repos WG hasn't voted to create the SIG (in fact the issue about it is still open: https://github.com/ossf/wg-securing-software-repos/issues/14), but AFAICT that work is a) ongoing regardless b) embodies a "SIG", at least in my opinion, and c) exists or should exist under that WG.

This is a good call-out! It's not clear to me if the Great Artifact Repository Audit is legacy or not, as that work started in parallel with this effort to update the SIG lifecycle doc.

Can the Security Software Repos WG decide if they're officially adopting the Great Artifact Repository Audit, or if not, notify them so they can seek another WG (or the TAC) as their governing body?

SecurityCRob (approved):
I am fine with these updates clarifying SIGs.

process/sig-lifecycle.md
```diff
@@ -1,13 +1,41 @@
 # Special Interest Group (SIG) Life Cycle
 
-Special Interest Groups are under the direct governance of a Working Group (WG) which can create and close them as necessary and appropriate. SIGs are expected to have a very specific goals and objectives at the time of their creation which may or may not define if they are terminable and is entirely dependent on the nature of the SIG.
+Special Interest Groups are bound to achieving a very specific goal. These groups may be terminated upon completion of their designating tasking or continued for larger and ongoing efforts. The creation of a SIG must dictate the focus, intent, goals, and deliverable(s) as appropriate. SIGs may report to a WG or directly to the TAC as their governing body.
 
-The lifecycle of SIGs is therefore minimal and as follows.
+It is expected that the primary output of a SIG is not software. If the primary output is software, the work should be organized as a [project](./project-lifecycle.md).
```
mlieberman85, line 5: Do we have any mechanisms for a SIG turning into a project? I have seen discussion in some of the groups that a body of work starts off mostly with a white paper or similar deliverable but over time there's interest in potentially extending it into code?

bobcallaway (approved):

process/sig-lifecycle.md
```diff
@@ -1,13 +1,41 @@
...
 
-## Active
+SIG process should be minimal, however we do need some process to at least ensure we have an accurate list of all the active SIGs in the OpenSSF. This document uses "must" to describe what items are required, "should" to suggest items that should be strongly considered (but not required), and "may" for suggested guidance.
```
mlieberman85, line 7: > SIG process should be minimal, however we do need some process to at least ensure we have an accurate list of all the active SIGs in the OpenSSF.

Why?

process/sig-lifecycle.md
```diff
@@ -1,13 +1,41 @@
...
+
+## To become `Active`:
+
+* A governing body must agree to govern the SIG
+  * The governing body may have its membership vote, with notice given at the meeting prior to the vote
+
+## Once `Active`:
+
+* The governing body must list information about the SIG on its README, including current state, chairs, and a statement of focus, intent, goals, and/or deliverables
+  * SIGs may have a repository (prefixed with `sig-`) that include the information listed above
```
lehors, line 23: I actually don't think that this is a good idea. I think we're better off not encoding in the repo name this kind of information that we might want to change in the future. For instance, if we hadn't had some repos with -wg in it we could have changed them to SIGs to be more aligned with other LF orgs. @mlieberman85 pointed out that a SIG might want to morph into a Project at some point. If we don't have sig- in the repo name this is just a matter of updating the README. If we have sig- in it it's more difficult to change.

process/sig-lifecycle.md
steiza: I'm not sure which part the "why" is directed at.

If it's the first part (the SIG process should be minimal) that's so there's less process friction and contributors can focus on their work and not navigating undue process.

In case the "why" is directed at the second part, we need a list of active SIGs to understand the activities of OpenSSF contributors for reporting on progress, ensuring the work is compatible with the OpenSSF's mission, etc.

process/sig-lifecycle.md
```diff
@@ -1,13 +1,41 @@
...
-## Inactive
+## To become `Tentative`:
 
-* The SIG is no longer active, either it has completed its mission or the mission was abandoned. At which state, the WG the SIG is under will determine a termination status and record it within the corresponding repository.
+* This is the default state of a new SIG that is not yet active
+
+## To become `Active`:
+
+* A governing body must agree to govern the SIG
+  * The governing body may have its membership vote, with notice given at the meeting prior to the vote
```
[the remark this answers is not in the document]
steiza: No, but this is good follow-on work!

PR #201, opened 2023-09-20
🔥 remove spelling workflow/action

ctcpip:
https://github.com/ossf/tac/pull/201#issuecomment-1727919565


jsoref:
Is there any feedback you have about the experience?

ctcpip:
spell checkers are great extensions to use on one's editor locally, but not something good to have as a status check on repositories

jsoref:
ok, thanks for trying it. I appreciate it.

ctcpip:
to elaborate:

- most spelling problems found are false positives
- nobody wants to see a red ❌ on a commit, especially for a spelling problem, and moreso when it's a false positive
- it is burdensome to constantly be updating configuration files / dictionaries
- doing the above increases surface area for false negatives

it would be more tolerable if it wasn't a status check, and instead did something like add a comment on the pull request that points out the spelling issues and line numbers (on newly added content only).  these comments could be addressed, or ignored if false positives, and it avoids most of the problematic issues with having it be a status check.

jsoref:
* It's possible to configure it as a [comment on PRs](https://github.com/check-spelling/spell-check-this/blob/3a35a1cec1e660ab04638cbd5bdd086cf2fcd5c5/.github/workflows/spelling.yml#L102-L117) (although that means it needs `pull-requests: write`, and some people don't like that)
* It's possible to have it not yield a :x: -- [`quit_without_error: true`](https://docs.check-spelling.dev/Configuration.html#quitwithouterror)
* In order to not frustrate people, you'd probably also want it to only check the changes for the PR, that's also possible: [`only_check_changed_files
: true`](https://docs.check-spelling.dev/Configuration.html#onlycheckchangedfiles)

Would you like to try that configuration?

jsoref:
Here's a demo https://github.com/check-spelling-sandbox/tac/pull/3#issuecomment-1728106994 of a configuration https://github.com/check-spelling-sandbox/tac/commit/ea13b38f46f8729925fc815a437dcf1db1659585 as described in 
https://github.com/ossf/tac/pull/201#issuecomment-1727940640

There are typos included in the PR base https://github.com/check-spelling-sandbox/tac/commit/b8154f4f3f942122e7862b79b3193c09460570c4, but they aren't reported in the PR.

If that's good enough, I can make a PR for it.

(Note that check-spelling doesn't have the ability to check the just the hunks that are added, although as it seems there's interest, I might prioritize that for the release after the one I'm about to ship.)

lehors:
I'm closing this for now while we try PR #203 which seems to address the main issue. We can reopen it later on if the experiment doesn't work.

lehors:
Having tested this for quite a while now it's become clear that while this checker catches a few typos here and there the number of false positive is so high that it is primarily a nuisance. It seems time to take it off this repo.

lehors (approved):
Let's do it.

bobcallaway:
Looks gewd to me

SecurityCRob (approved):

PR #202, opened 2023-09-20
Follow-on update to SIG process

steiza:
Based on feedback from https://github.com/ossf/tac/pull/182

### DIFF
--- a/process/sig-lifecycle.md
+++ b/process/sig-lifecycle.md
@@ -6,13 +6,9 @@ It is expected that the primary output of a SIG is not software. If the primary 
 
 SIG process should be minimal, however we do need some process to at least ensure we have an accurate list of all the active SIGs in the OpenSSF. This document uses "must" to describe what items are required, "should" to suggest items that should be strongly considered (but not required), and "may" for suggested guidance.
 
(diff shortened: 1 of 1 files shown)


process/sig-lifecycle.md
```diff
@@ -14,6 +14,9 @@ The SIG life cycle begins with interested contributors deciding to undertake thi
 
 ## To become `Active`:
```
steiza, line 16: I am purposely omitting a recommendation of a group meeting before being considered a SIG

process/sig-lifecycle.md
```diff
@@ -14,6 +14,9 @@ The SIG life cycle begins with interested contributors deciding to undertake thi
...
 
+* There must be a proposal of the focus, intent, goals, and/or deliverables of the SIG
+  * This could be in an Issue of the governing body's repository, a shared document, or in a README of the SIGs repository (if they have one)
+* The initial membership should be at least 3 individuals from 2 different organizations
```
steiza, line 19: This is a recommendation ("should") but of course is up for discussion

lehors (approved):
Next we'll need to specify what each stage entails, in terms of what a SIG gains from moving up onto the stage ladder, but this is definitely getting us closer to where we want to end up so I'm happy to merge this and go from there.
Thanks.

bobcallaway (approved):

PR #203, opened 2023-09-20
Tune check-spelling

jsoref:
This is an alternative to #201 as proposed in https://github.com/ossf/tac/pull/201#issuecomment-1727940640

* Add commenting on PRs
* Add commenting on push for repositories not in ossf
* Disable failure status reporting
* Only check changed files
  * Disable sarif reporting
  * Remove update job
* Adjust advice

With these changes, you'll get a comment like this: https://github.com/check-spelling-sandbox/tac/pull/3#issuecomment-1728284295 and no ❌.

### DIFF
--- a/.github/actions/spelling/advice.md
+++ b/.github/actions/spelling/advice.md
@@ -1,31 +1,4 @@
 <!-- See https://github.com/check-spelling/check-spelling/wiki/Configuration-Examples%3A-advice --> <!-- markdownlint-disable MD033 MD041 -->
-<details><summary>If the flagged items are :exploding_head: false positives</summary>
+If items reported here are typos that you've added, please fix them.
 
-If items relate to a ...
-* binary file (or some other file you wouldn't want to check at all).
-
-  Please add a file path to the `excludes.txt` file matching the containing file.
-
-  File paths are Perl 5 Regular Expressions - you can [test](
-https://www.regexplanet.com/advanced/perl/) yours before committing to verify it will match your files.
-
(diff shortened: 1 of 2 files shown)


steiza (approved):
I think it's nice to have the spelling feedback without blocking.

mlieberman85 (approved):
These all seem reasonable!

lehors (approved):
Ok, this sounds like a reasonable compromise. Let's try it.

lehors:
@jsoref I'm sorry but I don't understand how to parse this sentence. "To accept ✔️ these unrecognized words as correct and remove the previously acknowledged and now absent words, run the following commands".
First, I think the ✔️ symbol doesn't help, instead it breaks the sentence. But then I really struggle with that part: "remove the previously acknowledged and now absent words". When were words previously acknowledged? Why are they absent now?
I'm confused.

jsoref:
First thanks for your feedback, it really helps a lot.

The ✔️s are something I added in v0.0.20 (current is v0.0.21, although I was hoping to ship a v0.0.22 today...) -- https://github.com/check-spelling/check-spelling/commit/ccfe971d6b962fdd00905532bdeff5dab0fcc3a8 to try to make workflows easier to identify/discover.

Words in `expect.txt` (or similar) were accepted by previous commits.
They can be absent for a number of reasons:
1. they could have been removed
2. they could no longer be checked (`excludes.txt` / `only.txt`)
3. they could be masked (`patterns.txt`)
4. they could be in the dictionary (`extra_dictionaries` / `allow.txt`)
5. someone could have accidentally added a word to `expect.txt` that didn't belong there

PR #220, opened 2023-11-16
Add TAC Member R&Rs

afmarcum:
Proposed language for TAC Member R&Rs. This proposal pulls from Walli's P&Ps document and attempts to follow the WG Lead R&Rs format.

Open for feedback and patches. Thank you!

### DIFF
--- a/process/tac-member-R&Rs.md
+++ b/process/tac-member-R&Rs.md
@@ -7,7 +7,7 @@
 - Approve, establish, structure, organize, and archive Technical Initiatives and associated policies and procedures 
 - Provide sponsorship and guidance to new Technical Initiatives that petition to join or be created within OpenSSF
 - Participate in Technical Initiatives regularly to ensure alignment with the Technical Vision and MVSR of OpenSSF, and to identify any resource or funding requirements
-- Coordinate such other technical community matters related to the success of Technical Initiatives and the mission of the OpenSSF
+- Coordinate technical community matters related to the success of Technical Initiatives and the mission of the OpenSSF
 - TAC and Technical Initiative meetings shall be open, public meetings. For special circumstances, the TAC may hold meetings limited to the TAC voting representatives, invited guests, and LF staff
 - Annually elect a Chairperson who will preside over meetings, set the agenda for meetings, ensure meeting minutes are taken, and serve on the Governing Board as the TAC’s representative (the “TAC Representative”)
 - Time commitment: average 5+ hours per week
--- a/process/tac-member-R&Rs.md
+++ b/process/tac-member-R&Rs.md
@@ -1,8 +1,13 @@
 # TAC Member Roles & Responsibilities
 
 - Commit to abiding by the OpenSSF Code of Conduct
-- Regularly attend TAC meetings, participate in discussions, voting, and ad-hoc meetings as needed to support Technical Initiatives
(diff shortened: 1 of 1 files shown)


SecurityCRob (approved):

steiza (approved):

process/tac-member-R&Rs.md
```diff
@@ -0,0 +1,13 @@
+# TAC Member Roles & Responsibilities
+
+- Commit to abiding by the OpenSSF Code of Conduct
```
lehors, line 3: While I certainly agree with the statement it applies to everybody engaging in OpenSSF, not just the TAC. Maybe it's better to leave it off and keep this list to what is specific to TAC members?

process/tac-member-R&Rs.md
```diff
@@ -0,0 +1,13 @@
...
+- Provide ongoing, technology-neutral technical guidance and sponsorship to Technical Initiatives throughout the respective lifecycle stages
+- Establish and drive an overall Technical Vision aligned with the Foundation's MVSR for the community 
+- Regularly attend TAC meetings, participate in discussions, voting, and ad-hoc meetings as needed to support Technical Initiatives
+- Approve, establish, structure, organize, and archive Technical Initiatives and associated policies and procedures 
+- Provide sponsorship and guidance to new Technical Initiatives that petition to join or be created within OpenSSF
+- Participate in Technical Initiatives regularly to ensure alignment with the Technical Vision and MVSR of OpenSSF, and to identify any resource or funding requirements
+- Coordinate such other technical community matters related to the success of Technical Initiatives and the mission of the OpenSSF
```
lehors, line 10: "Coordinate such other"?? Something doesn't seem right here, unfortunately I'm not sure what the intent is so I can't really suggest a correction. Just remove "such"? But who is the coordination meant to be with?

lehors (changes requested):
Thanks for putting this together.

process/tac-member-R&Rs.md
afmarcum: Thanks for the catch, @lehors! I was mashing together the ideas of a few other folks and grammar lost this battle!

di (approved):

process/tac-member-R&Rs.md
afmarcum: @presidentoor do you want to weigh in here?

presidentoor: I think the language should read: "Abide by the OpenSSF Code of Conduct"

lehors: Ok but the question is whether we should have any such language here at all. I think not unless we go and add it everywhere else.

lehors: Ok but the question is whether we should have any such language here at all. I think not unless we are going to go back and add it everywhere else.

CoS-Harry: I don't know where everywhere else is as I view this document as foundational and standalone. This document resolves into the greater set of documents anyone may read about the TAC governance. The WG R&R should surely have the same note and I think promulgating the Code Conduct far and wide through these types of documents is useful and necessary.

SecurityCRob (approved):

lehors (approved):
LGTM. Thanks.

PR #227, opened 2023-11-28
Add removed Charter language for TAC operations

sevansdell:
Add removed charter language for TAC operations to the TAC Process & Procedures

### DIFF
--- a/process/README.md
+++ b/process/README.md
@@ -1,7 +1,7 @@
 
 # I. Overview
 
-This document describes the Open Source Security Foundation (OpenSSF) lifecycle process for Technical Initiatives (TI) which include Working Groups (WG), Projects (developing code or specifications), and Special Interest Groups (SIG).
+This document describes the Open Source Security Foundation (OpenSSF) lifecycle process for Technical Initiatives (TI) which include Working Groups (WG) (non-software focused), Projects (developing code or specifications), and Special Interest Groups (SIG). 
 
 The authority that governs this process is as follows:
 
--- a/process/README.md
+++ b/process/README.md
@@ -21,6 +21,8 @@ flowchart TD
 
 The process is designed to be flexible to enable a Project to move in and out of a Working Group as deemed appropriate by the TAC.
 
+A SIG can be developed within a WG or Project to address a specific need.
+
 # II. Lifecycle
 
 All Technical Initiatives follow a common lifecycle, with 4 stages: 
--- a/process/README.md
+++ b/process/README.md
@@ -1,7 +1,7 @@
 
 # I. Overview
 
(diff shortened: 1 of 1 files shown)


SecurityCRob (approved):

process/README.md
```diff
@@ -1,7 +1,7 @@
 
 # I. Overview
 
-This document describes the Open Source Security Foundation (OpenSSF) lifecycle process for Technical Initiatives (TI) which include Working Groups (WG), Projects (developing code or specifications), and Special Interest Groups (SIG).
+This document describes the Open Source Security Foundation (OpenSSF) lifecycle process for Technical Initiatives (TI) which include Working Groups (WG) (non-software focused), Projects (developing code or specifications), and Special Interest Groups (SIG).
```
lehors, line 4: First, given our change in the definition of projects, this should at least be:
```suggestion
This document describes the Open Source Security Foundation (OpenSSF) lifecycle process for Technical Initiatives (TI) which include Working Groups (WG) (non-software and non-specification focused), Projects (developing code or specifications), and Special Interest Groups (SIG). 
```
But I don't know that this is even correct because [the project lifecycle reads](https://github.com/ossf/tac/blob/main/process/project-lifecycle.md#sandbox-entry-requirements-and-considerations):

> It is expected that the initial code or specification developed by an OpenSSF WG be kept within their repository and will not function as a Project in its own right. Should the initial WG code or specification grow and mature that it warrants its own Project status, then it is subject to Sandbox entry requirements.

So, while the primary focus of a WG may not be to develop software or specs it can still do some of that so it doesn't seem to be a useful way to distinguish WGs.

sevansdell:
I understand – FYI We are proposing taking OUT of the charter how TAC defines Technical initiatives and only having those defined in the TAC Process and Procedures. These “definition type” items land only on the TAC Overview page.

As long as the wording meets the goals of the TAC, I am in favor of what you decide (I don’t have a strong opinion one way or the other about saying WGs are software focused.) This is a prompt to make sure we have the language TAC wants prior to the board meeting where the TAC TI definition is potentially removed from the charter.

Does that make sense? Can set up a quick call if needed.

Thanks,

Sarah



Internal Use - Confidential

lehors:
> I understand – FYI We are proposing taking OUT of the charter how TAC defines Technical initiatives and only having those defined in the TAC Process and Procedures. These “definition type” items land only on the TAC Overview page. As long as the wording meets the goals of the TAC, I am in favor of what you decide (I don’t have a strong opinion one way or the other about saying WGs are software focused.) This is a prompt to make sure we have the language TAC wants prior to the board meeting where the TAC TI definition is potentially removed from the charter. Does that make sense?

Yes, it does and I agree with the goal.

Issue #229, opened 2023-11-29
projects contribution template


hythloda:
We should have a template for projects to be added to the OpenSSF

lehors:
As described in the project lifecycle, our current [submission process](https://github.com/ossf/tac/blob/main/process/project-lifecycle.md#submission-process) is for people to submit a PR rather than open an issue (although they will need to also open an issue to track the legal clearance part), and we have templates for this, such as [the template to propose Sandbox project](process/templates/PROJECT_NAME_sandbox_stage.md).

We talked about possibly creating PR templates in GitHub but my past experience with that was rather disappointing. We probably should check whether the situation has improved though.

hythloda:
The goal of the issue template is to help make the process more discoverable. We have a lot of feedback that the current [submission process](https://github.com/ossf/tac/blob/main/process/project-lifecycle.md#submission-process) is too hard to find and they need to be experts in OpenSSF to submit a project, so this issue's goal is to walk them through all the steps more. 
 
They still need a PR like you @lehors mentions and that is one of the questions in the issue they submit.

Please see the draft issue template (still a work in progress and see if you like it. Would like edits in the yaml for improvements!

This is for [sandbox project contributions](https://github.com/ossf/tac/blob/project_contribution_template/.github/ISSUE_TEMPLATE/project_contribution_sandbox.yaml)

And [here for incubating](https://github.com/ossf/tac/blob/project_contribution_template/.github/ISSUE_TEMPLATE/project_contribution_incubate.yaml)

In the parser description are under the input, but other than that that is what the issue will look like.

Ironically just this morning the CNCF person got back to us on their methods so I incorporated that into this one, at least in the sandbox one saying they do similar. [You can see thiers here](https://github.com/cncf/sandbox/issues/new/choose).

david-a-wheeler:
Hi, I'm confused. We *DO* have an OpenSSF project template: https://github.com/ossf/project-template

I believe that by default, when a project is created on GitHub, those files are copied in (unless you *unselect* this template). I just added `SECURITY.md` to it, so there's a default `SECURITY.md` file. We should probably add more things to the template, e.g., maybe markdownlint or a universal linter.

lehors:
@david-a-wheeler My understanding is that this was meant to be for the request itself not for the repository.

hythloda:
[...]

PR #230, opened 2023-11-29
Project contribution template

hythloda:
I am unsure how to test this.

If this issue template works, then perhaps we need another issue template that is for creating TAC agenda items?

Any other issue templates needed?


hythloda:
Fixes #229

PR #252, opened 2024-01-24
Add CODEOWNERS

marcelamelara:
Adding a CODEOWNERS file for future PR notifications etc.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+* @bobcallaway @camaleon2016 @lehors @marcelamelara @mlieberman85 @SecurityCRob @sevansdell @steiza @torgo


mlieberman85 (approved):
LGTM

SecurityCRob (approved):
+1

marcelamelara:
The spell checker is unhappy about the word "codeowners". I can upload a file that makes the error go away, if folks think that's necessary.

mlieberman85:
I think we should turn off the spell check being a blocker.

lehors:
> I think we should turn off the spell check being a blocker.

Actually that was already done. We can ignore the errors and merge anyway.

Otherwise I must say that I'm torn about adding this file because it is yet one more piece of data to maintain but I understand the benefits so I'm not opposed to it.

marcelamelara:
> Otherwise I must say that I'm torn about adding this file because it is yet one more piece of data to maintain

Actually, if there's a Github team for the TAC, it may reduce overhead to add the team as the codeowner, rather than individuals in here. This way, whenever the team is updated when the TAC is updated, we still don't need to make changes to this file.

lehors (approved):
Thank you! This last commit addresses my concern.

steiza (approved):
Looks good to me!

Once this lands, we can check "Require review from Code Owners" in branch protection (https://github.com/ossf/tac/settings/branches). We could also consider migrating from branch protection to rulesets (https://github.com/ossf/tac/settings/rules).

PR #260, opened 2024-02-08
Group projects together

lehors:
Upon further review of the organizational structure overview graph I realized it fails to label the projects as such. It actually looks like sigstore is the only project we have!
This PR fixes that.
Some tuning is probably possible to prettify the graph a bit but let's get it right first.

### DIFF
--- a/organizational-structure-overview.md
+++ b/organizational-structure-overview.md
@@ -49,54 +49,73 @@ Legend:
 
 
 ```mermaid
-flowchart TB
-    A([Governing Board])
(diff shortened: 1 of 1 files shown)


mlieberman85 (approved):
lgtm

SecurityCRob (approved):

marcelamelara (approved):
thanks @lehors !

bobcallaway (approved):

steiza (approved):

PR #262, opened 2024-02-13
TAC sponsor should be sandbox stage requirement for WGs

marcelamelara:
The current WG lifecycle process docs don't currently require a TAC sponsor until _after_ the sandbox stage has been reached, even though the intent is for WGs the have TAC sponsor _before_ applying for sandbox status. This PR adjusts the WG lifecycle process to reflect that having a TAC sponsor should be a requirement for WGs when applying for the sandbox stage.

### DIFF
--- a/process/project-lifecycle.md
+++ b/process/project-lifecycle.md
@@ -44,7 +44,7 @@ The OpenSSF Sandbox is the entry point for early stage Projects and has four goa
 * Provides project updates to OpenSSF Marketing Committee as requested.
 
 #### Project Support
-* Receives guidance on technical direction from TAC.
+* Receives a TAC and WG sponsor for guidance on technical direction.
 * Receives consideration as in-scope for any submission to an OpenSSF-managed conference or event.
 * Receives OpenSSF Code of Conduct Committee support.
 * Reserved space for project updates in OpenSSF newsletters.
@@ -55,6 +55,10 @@ The OpenSSF Sandbox is the entry point for early stage Projects and has four goa
 
 * Projects must have a minimum of two maintainers with different organization affiliations.
 * Projects must be aligned with the OpenSSF mission _and_ either be a novel approach for existing areas or address an unfulfilled need. It is expected that the initial code or specification developed by an OpenSSF WG be kept within their repository and will not function as a Project in its own right. Should the initial WG code or specification grow and mature that it warrants its own Project status, then it is subject to Sandbox entry requirements. It is preferred that extensions of an existing OpenSSF project collaborate with the existing project rather than seek a new project.
+* Projects must seek one TAC sponsor and one WG sponsor (if reporting to a WG)
+    * TAC or WG sponsor agrees to attend Project meetings regularly
+    * TAC or WG sponsor does not need to have a formal role in Project, e.g., maintainer
+    * TAC or WG sponsor requests TAC approval
 * If contributing an existing project to the OpenSSF, the contribution must undergo license and IP due diligence by the Linux Foundation (LF).
 
 See [Submission Process](#submission-process) below and [Sandbox application](templates/PROJECT_NAME_sandbox_stage.md).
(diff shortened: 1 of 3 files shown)


marcelamelara:
I should have clarified this in the PR description. I posed an initial question about why the TAC sponsor (as currently documented) doesn't seem to be required for WGs before the incubating stage, even though I was under the impression that TAC sponsors were required sooner. A couple of folks replied that the intent apparently was to have a TAC sponsor be a requirement from the start, although the docs don't currently reflect this.

If this is open for debate, we can close this PR and open up an issue first.

CC @SecurityCRob and @steiza who had chimed in on Slack as well.

lehors (approved):
So, this actually appears to match what we did for the AI/ML WG, which was proposed and eventually approved with a sponsor last year. Based on that I'm in favor of this change.
Thanks!

steiza (approved):

marcelamelara:
Talking to some project leads, it looks like a TAC sponsor was also needed to accept recent projects like protobom and SBOMit as a sandbox project (within the Security Tooling WG) though the project lifecycle docs don't reflect this. So I'll be pushing some updates to align with our current process.

SecurityCRob (approved):

process/project-lifecycle.md
```diff
@@ -44,7 +44,7 @@ The OpenSSF Sandbox is the entry point for early stage Projects and has four goa
 * Provides project updates to OpenSSF Marketing Committee as requested.
 
 #### Project Support
-* Receives guidance on technical direction from TAC.
+* Receives a TAC and WG sponsor for guidance on technical direction.
```
steiza, line 47: This is a bit different than what we defined for SIGs, which are similar to Projects in that they can report directly to the TAC or to a WG.

You can see the details of that discussion on https://github.com/ossf/tac/pull/182, but where we landed was to define near the top of https://github.com/ossf/tac/blob/main/process/sig-lifecycle.md:

> SIGs may report to a WG or directly to the TAC as their governing body.

... and then after that refer to their governing body instead of WG or TAC.

There's definitely a balance here. On one hand, we don't want a TAC vote for everything or we risk the TAC being a bottleneck for the organization to get anything done. On the other hand, the way the process is currently set up, any change in lifecycle involves a pull request in the TAC repo, and it is nice to be at least informed of changes taking place across the org.

I would be in favor of WGs adopting Projects at the Sandbox stage without requiring TAC approval, but I'm open to further discussion!

marcelamelara: My initial reason for spelling it out like this is that several recent projects like SBOMit and protobom _have_ sought out a TAC sponsor to become sandbox. I totally agree that we shouldn't necessarily require TAC approval for every single TI either, so as you say, there's a balance to be struck. But the current TI Gives and Gets seems to make this a requirement for all TIs. So, some docs need to be adjusted, I'm just not sure where we want to draw the line.

steiza: I think the TI Gives + Gets is the odd one out here, so how about we do something like:

Change https://github.com/ossf/tac/blob/main/process/TI-Gives%2BGets.md?plain=1#L17:

> `* TI must find an aligned WG to host the TI or must have a TAC sponsor that can help guide the TI through processes.`

Could also consider `and/or`.

We could also consider changing https://github.com/ossf/tac/blob/main/process/TI-Gives%2BGets.md?plain=1#L27C2-L27C64:

> `* TI receives guidance on technical direction if they have a TAC sponsor`

I did a quick look at https://github.com/ossf/tac/tree/main/process/templates to ensure the Project and SIG templates don't explicitly require a TAC sponsor, so I think we're good there.

torgo (approved):

bobcallaway (approved):

process/project-lifecycle.md
marcelamelara: Thanks for these suggestions! What I like is that these will actually lower the barrier to entry for sandbox Projects and SIGs a little bit, so I'm definitely happy to adjust the TI Gives + Gets.

marcelamelara: How do we reconcile the need for opening a PR in this repo to submit a project's or SIG's sandbox application (which effectively amounts to a TAC approval) with the desire to make the TAC approval optional for projects and SIGs? Is this really about online vs offline TAC votes? This is something we ought to clarify in the docs.

SecurityCRob (approved):

PR #274, opened 2024-03-01
Update PROJECT_NAME_sandbox_stage.md

SecurityCRob:
Per the exemplar xls I shared with the TAC, here is what changes I feel need to be made to our template.  Feedback/edits please

### DIFF
--- a/process/templates/PROJECT_NAME_sandbox_stage.md
+++ b/process/templates/PROJECT_NAME_sandbox_stage.md
@@ -4,6 +4,10 @@
 The project must have a minimum of two maintainers with different organizational affiliations.
   * "name, affiliation, GitHub ID"
 
+### Existing TI alignment
+The project should be aligned with an existing OpenSSF Working Group, SIG, or Project (TI).  The project commits to providing the hosting TI quarterly updates on progress.
+  * "name of TI that will host this project"
+
 ### Mission of the project
 The project must be aligned with the OpenSSF mission and either be a novel approach for existing areas, address an unfulfilled need, or be initial code needed for OpenSSF WG work. It is preferred that extensions of existing OpenSSF projects collaborate with the existing project rather than seek a new project.
   * "description of the project mission"
@@ -15,11 +19,14 @@ When contributing an existing Project to the OpenSSF, the contribution must unde
 ### Project References
 The project should provide a list of existing resources with links to the repository, and if available, website, a roadmap, contributing guide, demos and walkthroughs, and any other material to showcase the existing breadth, maturity, and direction of the project.
 
-| Reference          | URL |
-|--------------------|-----|
-| Repo               |     |
-| Website            |     |
-| Contributing guide |     |
-| Roadmap            |     |
-| Demos              |     |
-| Other              |     |
+| Reference           | URL |
+|---------------------|-----|
+| Repo                |     |
+| Meeting Agenda      |     |
+| OSSF Calendar Entry |     |
+| Website             |     |
+| Contributing guide  |     |
+| Security.md         |     |
+| Roadmap             |     |
+| Demos               |     |
+| Other               |     |
--- a/process/templates/PROJECT_NAME_sandbox_stage.md
+++ b/process/templates/PROJECT_NAME_sandbox_stage.md
@@ -4,7 +4,7 @@
 The project must have a minimum of two maintainers with different organizational affiliations.
   * "name, affiliation, GitHub ID"
 
-### Existing TI alignment
+### Sponsor
 The project should be aligned with an existing OpenSSF Working Group, SIG, or Project (TI).  The project commits to providing the hosting TI quarterly updates on progress.
   * "name of TI that will host this project"
 
--- a/process/templates/PROJECT_NAME_sandbox_stage.md
+++ b/process/templates/PROJECT_NAME_sandbox_stage.md
@@ -5,7 +5,7 @@ The project must have a minimum of two maintainers with different organizational
   * "name, affiliation, GitHub ID"
 
 ### Sponsor
-The project should be aligned with an existing OpenSSF Working Group, SIG, or Project (TI).  The project commits to providing the hosting TI quarterly updates on progress.
+Most projects will be sponsored by an existing OpenSSF Working Group, although in some cases a project may be sponsored by the TAC. The project commits to providing the sponsor quarterly updates on progress.
   * "name of TI that will host this project"
 
 ### Mission of the project
--- a/process/templates/PROJECT_NAME_sandbox_stage.md
+++ b/process/templates/PROJECT_NAME_sandbox_stage.md
@@ -6,7 +6,7 @@ The project must have a minimum of two maintainers with different organizational
 
 ### Sponsor
 Most projects will be sponsored by an existing OpenSSF Working Group, although in some cases a project may be sponsored by the TAC. The project commits to providing the sponsor quarterly updates on progress.
-  * "name of TI that will host this project"
+  * "name of the project sponsor: either a Working Group or the TAC"
 
 ### Mission of the project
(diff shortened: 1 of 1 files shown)


mlieberman85 (approved):
lgtm

marcelamelara (approved):

camaleon2016:
Should there be a requirement for a LICENSE.md file?

process/templates/PROJECT_NAME_sandbox_stage.md
```diff
@@ -4,6 +4,10 @@
 The project must have a minimum of two maintainers with different organizational affiliations.
   * "name, affiliation, GitHub ID"
 
+### Existing TI alignment
+The project should be aligned with an existing OpenSSF Working Group, SIG, or Project (TI).  The project commits to providing the hosting TI quarterly updates on progress.
```
steiza, line 8:
```suggestion
Most projects will be sponsored by an existing OpenSSF Working Group, although in some cases a project may be sponsored by the TAC. The project commits to providing the sponsor quarterly updates on progress.
```

process/templates/PROJECT_NAME_sandbox_stage.md
```diff
@@ -4,6 +4,10 @@
...
+  * "name of TI that will host this project"
```
steiza, line 9:
```suggestion
  * "name of the project sponsor: either a Working Group or the TAC"
```

process/templates/PROJECT_NAME_sandbox_stage.md
steiza: I think the alignment part is covered by "Mission of the project" below, and it makes sense to have a section explicitly about the sponsor.

Note that this is different than what we list on https://github.com/ossf/tac/blob/main/process/project-lifecycle.md where we say "Projects must seek one TAC sponsor and one WG sponsor" - but I'm pretty sure we actually want that to say "Projects must seek one TAC sponsor or one WG sponsor" to align with https://github.com/ossf/tac/blob/main/process/sig-lifecycle.md.

... speaking of which on the SIG Lifecycle docs we call this a "governing body" not a "sponsor" like we do on the Project Lifecycle docs. We should probably also standardize this naming, but I think it makes sense to keep this Project Template consistent with the wording on the Project Lifecycle doc.

steiza (commented):
I think we should more explicitly reflect the sponsor wording in the project lifecycle doc.

steiza:
> Should there be a requirement for a LICENSE.md file?

It wasn't changed in this pull request, but the project template has an existing section "IP policy and licensing due diligence"

process/templates/PROJECT_NAME_sandbox_stage.md
steiza: Oh! I should add that I don't think TI is appropriate here because it TIs can include SIGs or other Projects, and I think we only want Projects reporting to the TAC or a WG, not other SIGs (or other Projects!)

camaleon2016:
What do we consider a "consistent release cadence" to be? I believe this needs to be less subjective. Are we thinking at least once a year? every couple of years?

camaleon2016:
Way may want to have them either point to a Specific SLSA Track in the framework, point to the use of a different framework and justify why. If the desire is to use the SLSA framework then the track used and adherence to the track should be discussed in previous stages. Consider feasibility stage and security requirements.

process/templates/PROJECT_NAME_sandbox_stage.md
```diff
@@ -4,6 +4,10 @@
 The project must have a minimum of two maintainers with different organizational affiliations.
   * "name, affiliation, GitHub ID"
 
+### Sponsor
+Most projects will be sponsored by an existing OpenSSF Working Group, although in some cases a project may be sponsored by the TAC. The project commits to providing the sponsor quarterly updates on progress.
+  * "name of the project sponsor: either a Working Group or the TAC"
+
```
lehors, line 10: I think this is using the term "sponsor" in a way that is different from what we've used it so far. Indeed, we've used the term to refer to a *person*. For instance, the project lifecycle states that a project "Receives a TAC and WG sponsor for guidance on technical direction."
This text uses "sponsor" to refer to the entity the project reports to.

steiza: This is exactly the question - I think we're trying to describe the reporting relationship. For SIGs we use the term "governing body" (which feels maybe too formal, but at least is clear). For projects I though we were using the term "sponsor" in the same way we use "governing body" for SIGs.

Maybe @SecurityCRob can clarify? Are we referring to the reporting relationship here? Or is a sponsor something else that is separate from the reporting relationship?

lehors: I never used it that way but maybe it's a matter of interpretation and we all have different perspectives on that one...
I too feel that "governing body" is a bit strong, that's why I phrased the above as "the entity the project reports to" instead. :-) It seems a bit softer to me.

torgo: How about "deliverable" as a better word for this - as in "projects are deliverables of working groups..."

SecurityCRob: > This is exactly the question - I think we're trying to describe the reporting relationship. For SIGs we use the term "governing body" (which feels maybe too formal, but at least is clear). For projects I though we were using the term "sponsor" in the same way we use "governing body" for SIGs.
> 
> Maybe @SecurityCRob can clarify? Are we referring to the reporting relationship here? Or is a sponsor something else that is separate from the reporting relationship?

Reporting relationship

process/templates/PROJECT_NAME_sandbox_stage.md
```diff
@@ -15,11 +19,14 @@ When contributing an existing Project to the OpenSSF, the contribution must unde
...
-| Website            |     |
-| Contributing guide |     |
-| Roadmap            |     |
-| Demos              |     |
-| Other              |     |
+| Reference           | URL |
+|---------------------|-----|
+| Repo                |     |
+| Meeting Agenda      |     |
+| OSSF Calendar Entry |     |
```
steiza, line 26:
```suggestion
```

These tables are hard! I can see why it'd be useful to have meeting agendas, if available, but after spending several hours reading these doc I *just now* realized that anything beyond the repository is optional. We don't reference an agenda / calendar meeting though, so I think it makes sense to remove it since it isn't clear if it's optional.

lehors: I agree. My initial intention with this stuff was to keep it as short as possible. My feeling was that a link to the main TI's page/repo should provide most of what we need. If it doesn't then we can then ask for additional info when we do the review.

steiza (approved):
LGTM!

process/templates/PROJECT_NAME_sandbox_stage.md
```diff
@@ -4,6 +4,10 @@
 The project must have a minimum of two maintainers with different organizational affiliations.
   * "name, affiliation, GitHub ID"
 
+### Sponsor
+Most projects will be sponsored by an existing OpenSSF Working Group, although in some cases a project may be sponsored by the TAC. The project commits to providing the sponsor quarterly updates on progress.
+  * "name of the project sponsor: either a Working Group or the TAC"
```
lehors, line 9: The requirement that they provide regular updates literally means that we are asking them to **report** to the WG or TAC! :-)
So, we shouldn't shy away from what this relationship means and I think we should just go with:
```suggestion
Most projects will report to an existing OpenSSF Working Group, although in some cases a project may report directly to the TAC. The project commits to providing quarterly updates on progress to the group they report to.
  * "name of the group the project reports to: either a Working Group or the TAC"
```

sevansdell (approved):

SecurityCRob:
5 of 9 TAC members have approved.  merging

PR #302, opened 2024-03-25
Correct spelling and process

omkhar:
Corrected a spelling mistake in the file name.
Clarified which staff can sign off on TI funding requests.
Clarified who can approve requests which are beyond the authority of the General Manager/delegate.
Made capitalization of 'OpenSSF Staff' consistent

Signed-off-by: Omkhar Arasaratnam <omkhar@linuxfoundation.org>

PR #303, opened 2024-03-26
Issue template for funding references process doc

hythloda:
This PR removes any reference to the process so that we have a single source of truth in the process documentation. 

The issue template was made before the process documentation but now that that is in place the issue does not need the information and it is best to reference the other location.

### DIFF
--- a/.github/ISSUE_TEMPLATE/funding_application.yml
+++ b/.github/ISSUE_TEMPLATE/funding_application.yml
@@ -14,7 +14,7 @@ body:
     attributes:
       value: |
         **Process:**
-        - Please review the process in the [TAC Process Documents](https://github.com/ossf/tac/blob/main/process/TI%20Funding%20Requst%20Process.md)
+        - Please review the process in the [TAC Process Documents](https://github.com/ossf/tac/blob/main/process/TI%20Funding%20Request%20Process.md)
   - type: markdown
     attributes:
       value: "Problem statement - This section aims to set the stage and boundaries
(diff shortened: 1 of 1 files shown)


lehors (approved):
I like it! Thanks. :-)

.github/ISSUE_TEMPLATE/funding_application.yml
```diff
@@ -13,55 +13,8 @@ body:
...
-          - If yes, recommend to support?
-            - How much funds will be available for other requests? (Should we approve one request for all the funds, or seek to approve multiple smaller requests)?
-          - If no, recommend to route through an alternate source?
-  - type: markdown
-    attributes:
-      value: |
-        **Communication of Final Decision:**
-        - Complete like a CFP. Notify in writing if it was accepted or not accepted. Put out a list of accepted proposals (like the events schedule).
+        **Process:**
+        - Please review the process in the [TAC Process Documents](https://github.com/ossf/tac/blob/main/process/TI%20Funding%20Requst%20Process.md)
```
bobcallaway, line 17: if/when https://github.com/ossf/tac/pull/302 merges, this link will need to be fixed (since the filename will change)

hythloda: Good call, let's keep this open until that closes.

marcelamelara (approved):
Thanks @hythloda !

SecurityCRob (approved):

bobcallaway (approved):

Issue #237, opened 2023-12-18
Make TAC voting process simpler and clearer


SecurityCRob:
The process for nomination, the voting process itself, the Governing Board nominated seats, how the TAC chooses a chair/vice-chair, as well as key dates/milestones should be clearly represented in the TAC voting process(1).  An example of a well-run process can be found with the Hyperledger TOC(2) that could be borrowed from as we update our documentation so that anyone can easily find, read, and understand how the election is run.  This work should be paired with Voter Eligibility clarity(3) that the GC has asked for.

(1) - https://github.com/ossf/tac/blob/main/elections/tac-and-scir-election-process.md
(2) - https://wiki.hyperledger.org/display/TSC/TOC+Election+2023
(3) - https://github.com/ossf/tac/issues/235

lehors:
I'm finally taking the time to look closer at this issue and I believe the following points need to be addressed:

* It is confusing for people to know whether they are eligible and registered to vote
* It is unclear where email related to the election are sent
* It is unclear how to handle people running for multiple seats: TAC "community seat", TAC "board seat", SCIR.
* The 2024 timeline did not respect the "Standard Election Timeline". There was no time for validation.
* The 2024 timeline did not specify the registration period (only specified the self nomination period which presumably ran concurrently).

Anything else?

SecurityCRob:
Agreed on your points, adding these:
- Process of determining how to fill vacant seats if someone steps down during their term (gb appointed and elected workflows)
- clearly documented timeline milestones 
- process of how TAC selects chair/vice-chair (internal vote, highest #s become chair and then vice)

sevansdell:
I think we are very close to a draft. No additional comments from me. Let's proceed with a process update in our documentation.

SecurityCRob:
has there been any progress on this?  can we close this out if it is complete?

lehors:
I believe PR #334 addressed this issue which I expected to get closed once the PR was merged. I suggest we just close it now.

sevansdell:
We have the election process going, and it is getting organized with great help from @Naomi-Wash. Closing out this issue. Great work.

PR #334, opened 2024-05-24
Improve election process and timeline

lehors:
This PR clarifies several aspects of the election process related to the TAC and SCIR positions, namely:

* Clarifies when and how the appointment by the GB takes place
* Expands on the timeline with clear milestones
* Specifies how the TAC chair and vice chair are elected
* Specifies what happens when a member steps down
* Specifies what happens when one runs for both TAC and SCIR
* Specifies that calls for nominations, registration, and vote need to be sent broadly to all TIs and general forums via email and Slack

This closes PR #237

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 # OpenSSF TAC Changelog
 
+* 2024-05-24: Improve election process and timeline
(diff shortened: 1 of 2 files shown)


SecurityCRob (approved):
thanks for putting this together for us @lehors !

bobcallaway (approved):
lgtm, thanks @lehors

PR #344, opened 2024-06-11
Iterate on funding application

steiza:
Arguably, our second funding cycle was a success with at least two requests approved!

When we went to evaluate the applications, it became clear that an exact funding amount was helpful for budgeting, and it wasn't clear that funding requests should come from a specific OpenSSF TI. These proposed changes aim to make the funding process more clear to applicants.


hythloda (approved):
awesome change!

bobcallaway (approved):

SecurityCRob (approved):
+1

marcelamelara (approved):
This is great, thanks @steiza ! Can you please add a changelog entry for this content update?

lehors (approved):
Yes!

PR #348, opened 2024-06-20
Add clarity wrt SIGs that need to be projects

mihaimaruseac:
When SIGs produce code or specification, they need to be projects, but this was not fully clear before. Aiming to clarify this.

### DIFF
--- a/process/Technical_Initiative_Lifecycle.md
+++ b/process/Technical_Initiative_Lifecycle.md
@@ -21,7 +21,7 @@ flowchart TD
 
 The process is designed to be flexible to enable a Project to move in and out of a Working Group as deemed appropriate by the TAC.
 
-A SIG can be developed within a WG or Project to address a specific need. Note that a SIG cannot produce code or specifications.
+A SIG can be developed within a WG or Project to address a specific need. Note that a SIG purpose is not to primary produce code or specifications (it can produce them as secondary goals, or as experimental POCs).
 
 # II. Lifecycle
 
--- a/process/Technical_Initiative_Lifecycle.md
+++ b/process/Technical_Initiative_Lifecycle.md
@@ -21,7 +21,7 @@ flowchart TD
 
 The process is designed to be flexible to enable a Project to move in and out of a Working Group as deemed appropriate by the TAC.
 
-A SIG can be developed within a WG or Project to address a specific need.
+A SIG can be developed within a WG or Project to address a specific need. Note that a SIG cannot produce code or specifications.
 
 # II. Lifecycle
 
--- a/process/Technical_Initiative_Lifecycle.md
+++ b/process/Technical_Initiative_Lifecycle.md
@@ -21,7 +21,7 @@ flowchart TD
 
 The process is designed to be flexible to enable a Project to move in and out of a Working Group as deemed appropriate by the TAC.
 
-A SIG can be developed within a WG or Project to address a specific need. Note that a SIG purpose is not to primary produce code or specifications (it can produce them as secondary goals, or as experimental POCs).
+A SIG can be developed within a WG or Project to address a specific need. Note that a SIG's primary purpose is not to produce code or specifications (it can produce them as secondary goals, or as experimental POCs).
 
 # II. Lifecycle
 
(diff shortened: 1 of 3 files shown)


marcelamelara:
@mihaimaruseac Thanks for your PR. While I agree that clarifications in the process are needed, the distinction between SIG and Project isn't strictly about code vs. no code. It's about scope, rather.

SIGs are meant to be more independently managed within the hosting TI, considerably narrower in scope, and have a shorter lifetime. But this doesn't preclude the SIG from producing code that is experimental or a PoC for a spec. On the other hand, if the SIG eventually intends to create production-quality software beyond PoC, that's when I think it makes sense to migrate to a Project designation a Project designation since the Project lifecycle includes more software-related requirements.

Again, I think our process docs would benefit from clarity around these details, but I don't think we should discourage SIGs from producing any code at all.

lehors:
I agree with @marcelamelara. I think the proposed wording is a bit too strong. Our [governance structure]9https://github.com/ossf/tac/blob/main/organizational-structure-overview.md) states that the "primary output" of a SIG is "not software". This implies that it could be a secondary output and clearly does not prohibit any software development.
We discussed cases where a SIG might develop some small tool to support its activities and that would be ok. So the proposed wording should be revised to align with that.
Thanks.

process/Technical_Initiative_Lifecycle.md
```diff
@@ -21,7 +21,7 @@ flowchart TD
 
 The process is designed to be flexible to enable a Project to move in and out of a Working Group as deemed appropriate by the TAC.
 
-A SIG can be developed within a WG or Project to address a specific need.
+A SIG can be developed within a WG or Project to address a specific need. Note that a SIG cannot produce code or specifications.
```
mihaimaruseac, line 24:
```suggestion
A SIG can be developed within a WG or Project to address a specific need. Note that a SIG purpose is not to primary produce code or specifications (it can produce them as secondary goals, or as experimental POCs).
```

process/sig-lifecycle.md
```diff
@@ -22,7 +22,7 @@ The SIG life cycle begins with interested contributors deciding to undertake thi
 * SIGs may have regular meetings separate from their governing body, if so:
   * They should appear on the OpenSSF calendar
   * They should have a document with upcoming agendas and notes from past meetings
-* If the SIG starts to produce code or specifications, they should consider becoming a [project](./project-lifecycle.md) instead
+* If the SIG starts to produce code or specifications, they must migrate to a [project](./project-lifecycle.md) instead
```
mihaimaruseac, line 25:
```suggestion
* If the SIG starts to produce code or specifications, they should consider becoming a [project](./project-lifecycle.md) instead, especially if this becomes the main scope of the SIG.
```

mihaimaruseac:
Thank you for reviewing. I've made some amendments to the wording to include the feedback.

SecurityCRob (approved):
+1 on revised changes

process/Technical_Initiative_Lifecycle.md
```diff
@@ -21,7 +21,7 @@ flowchart TD
 
 The process is designed to be flexible to enable a Project to move in and out of a Working Group as deemed appropriate by the TAC.
 
-A SIG can be developed within a WG or Project to address a specific need.
+A SIG can be developed within a WG or Project to address a specific need. Note that a SIG purpose is not to primary produce code or specifications (it can produce them as secondary goals, or as experimental POCs).
```
marcelamelara, line 24: Very minor nit.
```suggestion
A SIG can be developed within a WG or Project to address a specific need. Note that a SIG's primary purpose is not to produce code or specifications (it can produce them as secondary goals, or as experimental POCs).
```

marcelamelara (approved):
Thanks for the revisions @mihaimaruseac ! LGTM.

marcelamelara:
I also think we need a changelog entry for this one? Or should we consider this an editorial change?

marcelamelara:
Thanks so much @mihaimaruseac !

lehors (approved):
LGTM. Thanks!

steiza (approved):

mlieberman85 (approved):
lgtm

torgo (approved):
lgtm

bobcallaway (approved):

Issue #215, opened 2023-11-13
OpenSSF Security MVP


Danajoyluck:
We need to define the minimal security requirements/baseline for OpenSSF projects. The requirements will include projects at different stages of the lifecycle.

This would be an extension of issue [214 Universal adoption of scorecard and best practices](https://github.com/ossf/tac/issues/214)

The effort is to drive cross-Linux foundation security standards. This is one of the outcomes of Linux Foundation Member Summit. 

The draft baseline is here: https://docs.google.com/document/d/1-NBXdKvEJ9Wsh2i7lDNYven4fY9Bn6uvNJM5ySlMrdg/edit

SecurityCRob:
related to https://github.com/ossf/tac/issues/45

Danajoyluck:
Had a discussion with CRob about how to formalize the establish the baseline incrementally, by publishing baseline for sandbox first. Here are the advices from CROb
1. Publish the baseline content to a centralized location, GitHub would be ideal for public access. Currently it’s on Google drive
2. Incorporate the baseline into the operating model for the specific life cycle
3. Incorporate the baseline into the TAC lifecycle definition document
4. Reuse the this TAC issue to raise PR for point 2 & 3? 

Can I get the precise list of the documents  for the PR in point 4?

steiza:
I am definitely in favor of security baselines being part of our existing lifecycle docs that we have in https://github.com/ossf/tac/tree/main/process.

I think once we're broadly aligned on content, we should make pull requests to modify those docs. What I've learned in the past is it might make sense to start will one scoped pull request (like just adding security requirements to Sandbox stage), to align on phrasing / formatting / content before we progress to the other lifecycle stages.

sevansdell:
@mlieberman85 May we use GUAC as an initial test TI for the Security MVP, per Zach's recommendation above.

@Danajoyluck do you have any outstanding items from your comment above that you need to get started with a proof of concept using GUAC?

How can I help?

lehors:
I've reviewed the latest version of the proposed baseline and I think implementing it would really move us forward. I'm in favor of adopting it sooner rather than later. As always we can still fine tune it as we gain experience implementing it.

lehors:
By the way we will need to define a transition path to phase this in. We could start with a few pilot projects and progressively require existing TIs to implement the different levels of requirements they are expected to fulfill according to their lifecycle status.

puerco:
[...]

PR #353, opened 2024-07-10
Create security_baseline.md

Danajoyluck:
GitHub version of the security baseline for TAC review.


process/baseline/security_baseline.MD
```diff
@@ -0,0 +1,254 @@
...
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119/).
+
+## Background
+The initiative was one of the outcomes of the October 2023 Linux Foundation Member Summit. Making open source software more secure is one of the top priorities across the Linux Foundation. OpenSSF is leading the charge. 
+
+In the United States, open source software is used across all critical infrastructure sectors defined by CISA (Cybersecurity and Infrastructure Security Agency), for example, health care, defense, financial services, utilities, telecommunications, etc. Open source security directly impacts national security, economics and social stability. Enhancing open source security is imperative. NIST has published a [Secure Softeare Development Framework](https://csrc.nist.gov/Projects/ssdf) (SSDF) as a result of Executive Order (EO) 14028 on "[Improving the Nation's Cybersecurity](https://www.federalregister.gov/documents/2021/05/17/2021-10460/improving-the-nations-cybersecurity)".
```
lehors, line 15:
```suggestion
In the United States, open source software is used across all critical infrastructure sectors defined by CISA (Cybersecurity and Infrastructure Security Agency), for example, health care, defense, financial services, utilities, telecommunications, etc. Open source security directly impacts national security, economics and social stability. Enhancing open source security is imperative. NIST has published a [Secure Software Development Framework](https://csrc.nist.gov/Projects/ssdf) (SSDF) as a result of Executive Order (EO) 14028 on "[Improving the Nation's Cybersecurity](https://www.federalregister.gov/documents/2021/05/17/2021-10460/improving-the-nations-cybersecurity)".
```

process/baseline/security_baseline.MD
```diff
@@ -0,0 +1,254 @@
...
+    * Higher security: 3072 bits or 4096 bits
+  * ECDSA (Elliptic Curve Digital Signature Algorithm):
+    * Minimum recommended key length: 256 bits
+    * Higher security: 384 bits or 521 bits
+  * Ed25519:
+    * Fixed key length: 256 bits
+
+For data at rest, follow the instructions from the storage solutions for encryption practices
+
+### Curl and openssl Tips
```
lehors, line 235:
```suggestion
### Curl and OpenSSL Tips
```

process/baseline/security_baseline.MD
```diff
@@ -0,0 +1,254 @@
...
+      * Reuse existing OpenSSF guides and technologies without imposing new requirements.
+  * **Continuous Improvement**
+    * **Objective**: Establish a consistent set of objective security measures for all participating foundations and projects.
+    * **Approach**:
+      * Provide clear, implementable, and definitive guidelines for maintainers and contributors.
+      * Rely on every adopting project to submit issues to log the friction points and provide feedback to refine the baseline, facilitating easier adoption.
+  * **Documented Governance Process**
+    * **Objective**: Maintain and refine the baseline requirements effectively.
+    * **Approach**:
+      * Incoporate the baseline into OpenSSF Technical Advisory Council (TAC) [technical initiative life cycle process](https://github.com/ossf/tac/blob/main/process/project-lifecycle.md).
```
lehors, line 72:
```suggestion
      * Incorporate the baseline into OpenSSF Technical Advisory Council (TAC) [technical initiative life cycle process](https://github.com/ossf/tac/blob/main/process/project-lifecycle.md).
```

lehors:
LGTM but I have a couple of nits and a comment/question:
- I don't think we should add a `baseline` folder, just add the file in the `process` folder.
- the file extension should be `.md` rather than `.MD`

If, as discussed, we create a SIG to work on this, are we expecting them to work on this document in the TAC repo? That doesn't seem really desirable. Maybe they could create a copy of this document to work on in another repo. The instance in the TAC repo would then be refreshed from time to time when the SIG produces a new version that the TAC agrees to adopt. So, the instance in the TAC repo would remain the governing version while another copy exists in some other repo for the SIG to work on. Does that sound like a workable process? Any other idea?

mlieberman85:
> LGTM but I have a couple of nits and a comment/question:
> 
> * I don't think we should add a `baseline` folder, just add the file in the `process` folder.
> * the file extension should be `.md` rather than `.MD`
> 
> If, as discussed, we create a SIG to work on this, are we expecting them to work on this document in the TAC repo? That doesn't seem really desirable. Maybe they could create a copy of this document to work on in another repo. The instance in the TAC repo would then be refreshed from time to time when the SIG produces a new version that the TAC agrees to adopt. So, the instance in the TAC repo would remain the governing version while another copy exists in some other repo for the SIG to work on. Does that sound like a workable process? Any other idea?

I agree that we should have something external and only when read do we bring it into the TAC repo. I am making the assumption here that any material change to the baseline would require the TAC to approve.

lehors:
> I am making the assumption here that any material change to the baseline would require the TAC to approve.

Exactly. That's my assumption too. I don't think we want to delegate the responsibility of managing the baseline for the whole organization to the SIG.

david-a-wheeler:
Is this baseline *only* for the OpenSSF, or is the intention for this be a baseline applicable to *many* foundations? Perhaps at one time this was OpenSSF-only, but I think the longer goal is to create a "security baseline" that's useful for many foundations.

It's *fine* to put in the TAC repo in the short term, but I think in the long term it should *NOT* be in the TAC repo. It should instead eventually have its own repo or be part of the Best Practices WG repo (the latter would probably be easiest). I think that location will provide a stronger indication that this baseline can be used by other foundations.

Danajoyluck:
> Is this baseline _only_ for the OpenSSF, or is the intention for this be a baseline applicable to _many_ foundations? Perhaps at one time this was OpenSSF-only, but I think the longer goal is to create a "security baseline" that's useful for many foundations.
> 
> It's _fine_ to put in the TAC repo in the short term, but I think in the long term it should _NOT_ be in the TAC repo. It should instead eventually have its own repo or be part of the Best Practices WG repo (the latter would probably be easiest). I think that location will provide a stronger indication that this baseline can be used by other foundations.

The baseline is for OpenSSF. Can we put it here for now so that we can start updating the life cycle document and get the adoption going?

david-a-wheeler:
@lehors said:

> My take is that this is only meant for OpenSSF for now. When a SIG is launched with a goal to create one for the larger community they can create their own. Then, the TAC can decide how/when to revise this one.

Oh, I see. *This* PR is for an OpenSSF-specific baseline, and there will likely be a SIG created that starts with & builds on this material to create something for multiple foundations. Presumably, when that SIG is done, the TAC will decide if this PR's text will be replaced. That makes sense. Thanks so much for the clarification!

process/security_baseline.md
```diff
@@ -0,0 +1,253 @@
...
+
+In the United States, open source software is used across all critical infrastructure sectors defined by CISA (Cybersecurity and Infrastructure Security Agency), for example, health care, defense, financial services, utilities, telecommunications, etc. Open source security directly impacts national security, economics and social stability. Enhancing open source security is imperative. NIST has published a [Secure Software Development Framework](https://csrc.nist.gov/Projects/ssdf) (SSDF) as a result of Executive Order (EO) 14028 on "[Improving the Nation's Cybersecurity](https://www.federalregister.gov/documents/2021/05/17/2021-10460/improving-the-nations-cybersecurity)".
+
+In the European Union, [Cyber Resiliency Act](https://eur-lex.europa.eu/legal-content/EN/TXT/PDF/?uri=CONSIL:ST_17000_2023_INIT) (CRA) has declared a new role - Open Source Steward(Article 3(18a). The legislation is to ensure that consumers of products with digital features are protected, and manufacturers of these products are held accountable for product security. Open source security is the foundation for manufacturers to be compliant with the legislation.
```
marcelamelara, line 17: I don't know if this is how the CRA text is written, but the term "manufacturer" isn't typically used in the context of software. In SLSA, for example, the term we have agreed upon is the "software producer", so I would recommend using a term like "software producer" whenever referring to the entity that creates a software product.

process/security_baseline.md
```diff
@@ -0,0 +1,253 @@
...
+
+In the European Union, [Cyber Resiliency Act](https://eur-lex.europa.eu/legal-content/EN/TXT/PDF/?uri=CONSIL:ST_17000_2023_INIT) (CRA) has declared a new role - Open Source Steward(Article 3(18a). The legislation is to ensure that consumers of products with digital features are protected, and manufacturers of these products are held accountable for product security. Open source security is the foundation for manufacturers to be compliant with the legislation.     
+
+## Success Criteria
+The success of the security baseline SHOULD be quantified and qualified in a few dimension:
```
marcelamelara, line 20: nit
```suggestion
The success of the security baseline SHOULD be quantified and qualified in a few dimensions:
```

process/security_baseline.md
```diff
@@ -0,0 +1,253 @@
...
+ * **Security Outcomes Adoption and Advancement**
```
marcelamelara, line 21: This might just be my nitpicking, but it wasn't clear to me what you meant by "security outcomes" at first. To me, that sounds like this is referring to the improvement of the overall security posture, but this success criterion is really about adopting and advancing tooling/frameworks/etc. I suggest renaming this criterion to something that more clearly conveys the object and approach:

```suggestion
 * **Adoption and Advancement of Security Technologies**
```

process/security_baseline.md
```diff
@@ -0,0 +1,253 @@
...
+   * **Objective**: The security baseline promotes improved open source security outcomes.
+   * **Approach**: Utilize technical outcomes from the OSS community to implement, and verify the security baseline.
+   * **Collaborations**: Partner with other foundations to enhance OSS security through education, specifications, frameworks, and tooling.
+   * **Automation and Tooling**: Identify opportunities for automation and tooling to ease the burden on maintainers and contributors, facilitating secure by design and secure by default principles.
```
marcelamelara, line 25: As for the others, are there specific metrics that could be used as measures of success here?

process/security_baseline.md
```diff
@@ -0,0 +1,253 @@
...
+ * **Increased Adoption Rate**
```
marcelamelara, line 26: Just to make the distinction with the previous criterion clear, I suggest a slight phrasing change here as well:
```suggestion
 * **Increased Security Baseline Adoption Rate**
```

process/security_baseline.md
```diff
@@ -0,0 +1,253 @@
...
+   * **Objective**: Ensure the security baseline is widely adopted.
+   * **Metrics**:
+     * Three OpenSSF software-based pilot projects adopt the baseline by 9/15/2024.
+     * All OpenSSF software-based projects adopt the baseline by the end of 2024.
+     * At least two other LF foundations adopt the baseline by the end of 2024.
```
marcelamelara, line 31: What do we mean by "adoption" here? Is it enough for the other foundations to agree to the adoption, or is the expectation that they would have a foundation-wide implementation of the security baseline by the end of 2024? The former seems doable by the end of the year, the latter would be more challenging.

Also, I believe it was mentioned this one was a stretch goal for 2024. I would label it as such.

process/security_baseline.md
```diff
@@ -0,0 +1,253 @@
...
+* **Reduction in Security Findings**
+  * **Objective**: Measure the effectiveness of the baseline in improving a project’s security posture.
+  * **Metrics**:
+    * Use Scorecard statistics to track the reduction of Critical, High, and Medium risk findings through the project lifecycle. Benchmark is established at the adoption time. 
+    * Monitor the number of exceptions at both project and foundation levels, expecting a decline as projects mature and become more secure.
+* **Improved OSS Maintainer and Consumer Experience**
```
marcelamelara, line 37: Similar to my comment above, can we come up with a metric or two that would allow us to measure that maintainer/consumer experience has improved? I know this is incredibly hard, but finding some key measurable results may help in addressing future gaps etc.

marcelamelara: On the other hand, if we don't intend to measure this, I can see this point fitting under operating principles as well.

process/security_baseline.md
```diff
@@ -0,0 +1,253 @@
...
+**Shortage of contributors for tooling development**:  Security in open source can only be achieved at scale with the right tool. Lack of easily adopted tools prevents us from achieving a higher level of security baseline. 
+
+**Maintainers' time strained by disproportionate open-source consumer demands**: The security baseline aims to enable more secure software development at speed. Nevertheless, the primary constraint remains the limited time available to maintainers, coupled with the demand for open-source software maintenance, including the enhancement of software security.
+
+## Basic Operating Principles 
+To navigate these constraints, the following operating principles are adopted:
+
+  * **Strong Bias Towards Automation and Automatability**
+    * **Objective**: Enhance security by default and position security as an enabler rather than an inhibitor.
+    * **Approach**: Prioritize automation and automatability in all processes.
```
marcelamelara, line 57: This makes sense as an objective, but the approach, i.e. how we plan/recommend to prioritize and achieve automation/automatability, is not. Is there a more concrete approach we could describe here? Are there existing best practices this document could refer to?

process/security_baseline.md
```diff
@@ -0,0 +1,253 @@
...
+  * **Minimal, Achievable, and Practical Baseline Requirements**
+    * **Objective**: Design a security baseline that balances software reliability, performance, cost, and security.
+    * **Approach**:
+      * Ensure the baseline is minimal and achievable with current technology.
+      * Allow for incremental adoption throughout a software project’s lifecycle by shifting security left in the SDLC process.
+      * Reuse existing OpenSSF guides and technologies without imposing new requirements.
```
marcelamelara, line 63: I'm a little unclear as to what these two approaches really mean. 
* How does shift-left lead to incremental adoption? 
* With "Reuse existing guides/technologies without imposing new requirements", is the idea here that these guides won't change significantly adding more requirements? The reason I'm also a little confused here is that the security baseline itself _will_ likely impose new requirements on projects, so clarifying what "new requirements" are meant here would be helpful.

process/security_baseline.md
```diff
@@ -0,0 +1,253 @@
...
+  * **Continuous Improvement**
```
marcelamelara, line 64: Is this principle about continuous improvement of the security baseline? This was not immediately clear in the Objective and Approach.

process/security_baseline.md
```diff
@@ -0,0 +1,253 @@
...
+    * **Objective**: Establish a consistent set of objective security measures for all participating foundations and projects.
+    * **Approach**:
+      * Provide clear, implementable, and definitive guidelines for maintainers and contributors.
+      * Rely on every adopting project to submit issues to log the friction points and provide feedback to refine the baseline, facilitating easier adoption.
+  * **Documented Governance Process**
+    * **Objective**: Maintain and refine the baseline requirements effectively.
```
marcelamelara, line 70: This objective sounds a like the the previous operating principle. I would rephrase this to make the governance-related objective clearer.

marcelamelara (changes requested):
Thanks for submitting this PR @Danajoyluck ! I've reviewed the first half of the document, and I like the direction this is going. I do have several comments around clarity and specificity that will hopefully help improve this document.

process/security_baseline.md
Danajoyluck: > I don't know if this is how the CRA text is written, but the term "manufacturer" isn't typically used in the context of software. In SLSA, for example, the term we have agreed upon is the "software producer", so I would recommend using a term like "software producer" whenever referring to the entity that creates a software product.

Thanks @marcelamelara I will use software producer in other places. In CRA context, manufacturer needs to be used. I had to update my presentation to replace producer with manufacturer when talking about OpenSSF in light of CRA

process/security_baseline.md
Danajoyluck: @marcelamelara can we solicit feedback from maintainers during adoption about experience improvement measure? will a survey by the end of the adoption piloting be helpful?

process/security_baseline.md
Danajoyluck: @marcelamelara we will start tracking the adoption once it starts, through GitHub issues. The tracking includes adoption friction points, additional tooling needs, gaps in the baseline. This is echoing the same sentiment as security tool belt, how we can make security adoption easier through tooling.

process/security_baseline.md
Danajoyluck: @marcelamelara 
How does shift-left lead to incremental adoption? we embed security into the SDLC process, that's shifting security to the left. At various life cycle, the security risks of the projects are different, only the necessary security measures are taken into account for that life cycle, and build it up gradually as the software becomes more mature, that's the incremental approach supported by the shifting security to the left. What is a better way to represent this? 

Around "Reuse existing guides/technologies without imposing new requirements", the baseline aggregates the best security practices and technologies that are essential for developing secure software. I think we need to have a discussion about how we evolve the guides and the security baseline in harmony. How about we rephrase it to "Reuse existing guides/technologies with minimal new requirements"?

process/security_baseline.md
Danajoyluck: the three bullets under "Minimal, Achievable, and Practical Baseline Requirements" meant to ensure that the baseline is not adding additional burdens to the maintainers while adopting the baseline. The continuous improvement is more focused on  we reduce security burden on maintainers through continuous improvements,

process/security_baseline.md
```diff
@@ -0,0 +1,253 @@
...
+|Data in transit must be protected by cryptographic means.|To protect data from unauthorized access,  ensure data integrity, confidentiality and availability.|[Encrypting Data in Transit and at Rest](#Encrypting-Data-in-Transit-and-at-Rest) provides detailed rationale behind the baseline and implementation guidance.|[Use curl and openssl to verify](https://docs.google.com/document/d/1-NBXdKvEJ9Wsh2i7lDNYven4fY9Bn6uvNJM5ySlMrdg/edit#heading=h.59ofury7nggp) the TLS protocol version, cipher and certificate chain for HTTPS traffic.|
```
lumjjb, line 111: We had an issue with this in this during the prometheus security review in CNCF. I think the question here is depending on the model... E.g. if you run a service on istio, you would have TLS, but you wouldn't want the servers to actually run TLS themselves too... This is a weird one because its subjective on deployment and threat model. I would suggest against making this a requirement.

process/security_baseline.md
```diff
@@ -0,0 +1,253 @@
...
...able  vulnerabilities MUST be fixed promptly.|To identify vulnerabilities in the codebase you develop, and remediate critical vulnerabilities that are exploitable, ensure your project is adoptable.|Adopt a Static Application Security Testing (SAST) tool to run static code analysis, e.g. CI/CD pipeline. The code analysis runs against the codebase that you develop, identify vulnerabilities and provide recommendations on remediation.<br /><br />Consider [Enable CodeQL](https://github.com/github/codeql-action#usage) in your CI/CD pipeline.<br /><br />Examples:<br />OpenSSF:[Sigstore Fulcio](https://github.com/sigstore/fulcio?tab=readme-ov-file), [Scorecard](https://github.com/ossf/scorecard/tree/49c0eed3a423f00c872b5c3c9f1bbca9e8aae799)|[Scorecard checks SAST](https://github.com/ossf/scorecard/blob/7ce8609469289d5f3b1bf5ee3122f42b4e3054fb/docs/checks.md#sast)is used in your CI/CD and reports the finding through the vulnerabilities check. You SHOULD see the report reflects the vulnerability fixes.<br /><br />Scorecard MAY not detect some SAST tools.<br /><br />Example report: [Scorecard](https://scorecard.dev/viewer/?uri=github.com/ossf/scorecard&sort_by=risk-level&sort_direction=desc)|
```
lumjjb, line 109: Would this include publishing of VEX statements and ensuring that scorecard is able to interpret the VEX formats as well?

sevansdell (approved):
Thanks for requested updates.

process/security_baseline.md
marcelamelara: Interesting, thanks for clarifying.

process/security_baseline.md
```diff
@@ -0,0 +1,252 @@
...
...is published, maintained and followed.|Sets clear guidelines for selecting and maintaining secure dependencies.|Follow [Concise Guide for Evaluating Open Source Software](https://best.openssf.org/Concise-Guide-for-Evaluating-Open-Source-Software) to evaluate the  dependencies before using them in the project.<br /><br />Publish a dependencies policy to guide contributors on dependency management, using a stand-alone file or CONTRIBUTING.md.<br /><br />Example dependency policy:<br />CNCF: [Kubescape](https://github.com/kubescape/kubescape/blob/master/docs/environment-dependencies-policy.md), [Argo Helm](https://github.com/argoproj/argo-helm/blob/main/CONTRIBUTING.md#new-application-versions)<br /><br />The policy SHALL be added to SECURITY_INSIGHTS.yml section “dependencies” > “env-dependencies-policy”.<br /><br /> Example SECURITY_INSIGHTS.yml with dependencies policy: <br />CNCF: [Kubescape](https://github.com/kubescape/kubescape/blob/master/SECURITY-INSIGHTS.yml), [capsule](https://github.com/projectcapsule/capsule/blob/main/SECURITY-INSIGHTS.yml).|SECURITY_INSIGHTS.yml identifies the dependencies policy,  the policy provides guidance on dependencies evaluation and maintenance.|
```
marcelamelara, line 97: These table entries are quite difficult to read because of all of the in-lined links. Can you please put all URLs at the bottom of the document to make this table a bit clearer?

process/security_baseline.md
```diff
@@ -0,0 +1,252 @@
...
+
+## Security Baseline 
+As a software project progresses through the [OpenSSF technical initiative life cycle](https://github.com/ossf/tac/blob/main/process/project-lifecycle.md), it is beneficial to incrementally integrate more security measures.
+  * Sandbox stage - baseline awareness, and initial requirements
+  * Incubation stage - sandbox requirements plus additional baselines to sustain the secure project development and protect early adopters. 
+  * Graduated stage: Incubation baseline plus additional baselines to prepare for its general availability and sustainability. 
+
+This phased approach intends to support maintainers, contributors, and the community in innovating quickly with security built into the design or enabled by default.
+
+### Baseline - Once Sandbox
```
marcelamelara, line 82: I think it would be helpful to add a short summary of the overall goal of the Once Sandbox baseline, i.e. how do memory-safe languages and 2FA help achieve that goal?

process/security_baseline.md
```diff
@@ -0,0 +1,252 @@
...
+
+### Baseline - To Become Incubating
+
+As the project codebase grows and more features are added, increasing complexity, it becomes crucial to leverage security tools to identify vulnerabilities in the codebase or dependent software early on. Addressing critical issues early prevents costly fixes in the future. At this stage, projects SHOULD onboard to OpenSSF Scorecard by following the [installation instructions](https://github.com/ossf/scorecard-action#installation) of [Scorecard GitHub Action](https://github.com/ossf/scorecard-action). The Action runs on any repository change and raises alerts. ​​Repository administrators, organization owners, and people with write or maintain access to a repository can view the alerts in the repository’s Security tab. Ensure Scorecard is enabled for the project by following [Scorecard Verify Runs](https://github.com/ossf/scorecard-action?tab=readme-ov-file#verify-runs) instruction.
```
marcelamelara, line 90: This says "projects SHOULD onboard Scorecard", but it seems like there are one or two required Scorecard metrics below. [Current TI incubating requirements](https://github.com/ossf/tac/blob/main/process/TI-Gives%2BGets.md#givesrequirements-1) already include Scorecard as a MUST, so I think the baseline ought to be consistent with that and also state Scorecard onboarding as a MUST. The table below can then clarify which specific Scorecard metrics are needed for incubating.

process/security_baseline.md
```diff
@@ -0,0 +1,252 @@
...
+
+### Baseline - Once Incubating
+As a project matures and enters into the incubation stage, the project is expected to have more adopters. Additional security needs to be built into the project in order to sustain the secure project development and protect the consumers of the open source project.
```
marcelamelara, line 101: As above, what's the overall security goal of this baseline level?

process/security_baseline.md
marcelamelara: +1

marcelamelara (commented):
Thanks for the updates @Danajoyluck ! I left some specific comments, but have two high-level asks: 1) Can you please move the in-lined links in the security baseline tables to the bottom of the document? The actual text in the tables is currently quite hard to read and review. 
2) To help motivate adoption and provide clear reasoning for each baseline level, I suggest stating the overall security goal for each level. For example, at "Once Sandbox" the overall goal is to avoid some of the most common security issues from the start. This will also help explain why each requirement was chosen for that particular level.

Danajoyluck:
> Thanks for the updates @Danajoyluck ! I left some specific comments, but have two high-level asks: 1) Can you please move the in-lined links in the security baseline tables to the bottom of the document? The actual text in the tables is currently quite hard to read and review. 2) To help motivate adoption and provide clear reasoning for each baseline level, I suggest stating the overall security goal for each level. For example, at "Once Sandbox" the overall goal is to avoid some of the most common security issues from the start. This will also help explain why each requirement was chosen for that particular level.

Thanks a lot @marcelamelara for taking the time.

It will be very hard for maintainers to read the document if the links are not inline. Will this file view help with the review? https://github.com/ossf/tac/blob/22639387114e387c5f3f303c0d30ad88e9a79c1e/process/security_baseline.md

Around adoption, at least three projects have agreed to adopt. can we take an agile approach on this? I put the reasons why we need the baseline for each "level", would you consider that as goals?

SecurityCRob (approved):
I approve

process/security_baseline.md
```diff
@@ -0,0 +1,254 @@
...
+The success of the security baseline SHOULD be quantified and qualified in a few dimensions:
+ * **Adoption and Advancement of Security Technologies**
+   * **Objective**: The Security Baseline promotes improved open source security outcomes.
+   * **Approach**: Utilize technical outcomes from the OSS community to implement, and verify the security baseline.
+   * **Collaborations**: Partner with other foundations to enhance OSS security through education, specifications, frameworks, and tooling.
+   * **Automation and Tooling**: Identify opportunities for automation and tooling to ease the burden on maintainers and contributors, facilitating secure by design and secure by default principles.
+ * **Increased Security Baseline Adoption Rate**
+   * **Objective**: Ensure the Security Baseline is widely adopted.
+   * **Metrics**:
+     * Three OpenSSF software-based pilot projects meet the baseline requirements for each project's life cycle by 9/15/2024.
```
puerco, line 29: This sounds reasonable, as mentioned on https://github.com/ossf/tac/issues/215 both @protobom and @openvex will start piloting the baseline as soon as it is merged. Especially protobom is still working on setting its new org, so its an ideal time for us.

mlieberman85 (approved):
LGTM, there's a few caveats but as mentioned in chat I think we are at the point where we can iterate in the SIG and among the pilot projects like GUAC, OpenVEX, protobom, etc.

SecurityCRob (approved):
I look forward to seeing how our pilot projects work through achieving the baseline!

lehors (approved):
LGTM

torgo (approved):
lgtm

sevansdell (approved):

bobcallaway (approved):

SecurityCRob (approved):

process/security_baseline.md
```diff
@@ -0,0 +1,254 @@
...
...ities to enhance supply chain security, in both human and  machine processable format.|Refer to [OpenSSF SBOM Everywhere SIG](https://github.com/ossf/sbom-everywhere) for the importance of  SBOM generation, consumption and process improvement.<br /><br />Generate SBOM’s in your project’s GitHUb release workflow, ensure [minimal data fields recommended by NTIA](https://www.ntia.gov/report/2021/minimum-elements-software-bill-materials-sbom) are populated.<br /><br />Example SBOM: <br />OpenSSF: [GUAC](https://github.com/guacsec/guac/blob/main/.github/workflows/release.yaml) SPDX<br />CNCF: [Kyverno](https://github.com/kyverno/kyverno/blob/68df5af40e9820633162e8d1c0b9966032eb7eb8/.github/actions/publish-image/action.yaml#L3) CycloneDX<br /><br />SECURITY_INSIGHTS.yml SHALL be updated under “dependencies” > “sbom” to include all the build SBOM’s.<br /><br />[TAC project lifecycle governance process](https://github.com/ossf/tac/blob/main/process/project-lifecycle.md) SHALL be used for exception approval when there is no tooling support to implement the baseline.|SECURITY_INSIGHTS.yml identifies the SBOM’s. SMOB’s provide information for dependency management, vulnerability management. |
```
steiza, line 124:
```suggestion
|If release artifacts are created, SBOM is generated and distributed as a release artifact.|To provide accurate and complete inventory of the software version, license,  dependencies and vulnerabilities to enhance supply chain security, in both human and  machine processable format.|Refer to [OpenSSF SBOM Everywhere SIG](https://github.com/ossf/sbom-everywhere) for the importance of  SBOM generation, consumption and process improvement.<br /><br />Generate SBOM’s in your project’s GitHub release workflow, ensure [minimal data fields recommended by NTIA](https://www.ntia.gov/report/2021/minimum-elements-software-bill-materials-sbom) are populated.<br /><br />Example SBOM: <br />OpenSSF: [GUAC](https://github.com/guacsec/guac/blob/main/.github/workflows/release.yaml) SPDX<br />CNCF: [Kyverno](https://github.com/kyverno/kyverno/blob/68df5af40e9820633162e8d1c0b9966032eb7eb8/.github/actions/publish-image/action.yaml#L3) CycloneDX<br /><br />SECURITY_INSIGHTS.yml SHALL be updated under “dependencies” > “sbom” to include all the build SBOM’s.<br /><br />[TAC project lifecycle governance process](https://github.com/ossf/tac/blob/main/process/project-lifecycle.md) SHALL be used for exception approval when there is no tooling support to implement the baseline.|SECURITY_INSIGHTS.yml identifies the SBOM’s. SMOB’s provide information for dependency management, vulnerability management. |
```

mlieberman85 (approved):
LGTM

process/security_baseline.md
```diff
@@ -0,0 +1,254 @@
...
...t familiar with SLSA specification and available tooling, visit this [workshop material](https://github.com/slsa-framework/oss-na24-slsa-workshop?tab=readme-ov-file). <br /><br />SECURITY_INSIGHTS.yml SHALL be updated under “security-artifacts” > “other-artifacts” to include attestation information.<br /><br />Examples:<br />OpenSSF: [Sigstore Fulcio](https://github.com/sigstore/fulcio/releases), [Scorecard](https://github.com/ossf/scorecard/blob/main/.github/workflows/slsa-goreleaser.yml) |[Scorecard Signed-Releases Check](https://github.com/ossf/scorecard/blob/7ce8609469289d5f3b1bf5ee3122f42b4e3054fb/docs/checks.md#signed-releases) reports attestation is generated as a release artifact. When the Scorecard score is > 8 and !=10, [Allstar](https://github.com/ossf/allstar) will raise an issue and add new comments every 24 hours until the issue is resolved.<br /><br />SECURITY_INSIGHTS.yml identifies  the attestation file. The attestation is  verifiable using these [instructions](https://docs.github.com/en/actions/security-guides/using-artifact-attestations-and-reusable-workflows-to-achieve-slsa-v1-build-level-3#step-2-verifying-artifact-attestations-built-with-a-reusable-workflow).|
```
steiza, line 126: Minor: consider SLSA Build v1.0 Level 2+

In working with people to adopt SLSA Build Level 3, it's not always clear what that entails. My interpretation is that when you **verify** the provenance, you are checking that the workflow matches one you've vetted / audited. Of course, the project **producing** the provenance won't necessarily control how other people verify it. It's not (in my opinion) as simple as saying "use reusable workflows", as you don't actually get the SLSA Build Level 3 benefit unless you're verifying which workflow - reusable or not, at a specific revision - was used in the build.

process/security_baseline.md
```diff
@@ -0,0 +1,254 @@
...
+|Monitoring of security events is implemented if your project provides internet or infrastructure service on behalf of the foundation.|To monitor security relevant events for incident response.|If the project provides a service, monitoring SHALL be in place to raise actionable alerts when security relevant events meets pre-defined  thresholds, for example host level firewall configuration is changed.|Manual review.|
+|If your project provides internet or infrastructure service on behalf of the foundation, an initial  security audit is conducted. Audit findings are addressed.|To identify and remediate the vulnerabilities in the internet service.|Security audit SHALL be funded through the [TAC TI funding process](https://github.com/ossf/tac/blob/main/process/TI%20Funding%20Request%20Process.md).<br /><br />SECURITY_INSIGHTS.yml SHALL be updated under “security-assessments” with a link to the audit report.|SECURITY_INSIGHTS.yml identifies the security audit report. The report provides details of the audit methodology, findings and recommendations.|
+
+### Security Baseline - Graduated
```
steiza, line 131:
```suggestion
### Security Baseline - Once Graduated
```

steiza (approved):
I have a bunch of minor, nit-picky feedback. Great job synthesizing this complex topic, and I can't wait to get feedback from projects as we roll this out!

torgo (approved):
lgtm

marcelamelara (approved):
Thanks for the updates @Danajoyluck ! There are still some edits I'd like to see eventually, but I think this is ready for us and the SIG to begin iterating over.

PR #376, opened 2024-08-27
[security_baseline.md] Alter "or" to "to" to denote a range

fvsamson:
Only NIST provides parameters for a 521 bit elliptic curve, others (e.g. ECC Brainpool) provide parameters for a 512 bit curve instead.


fvsamson:
Alternatively this could be phrased:

    * Higher security: 384, 512 or 521 bits

SecurityCRob (approved):
+1

marcelamelara (approved):
LGTM, thanks for catching this!

lehors (approved):

PR #382, opened 2024-09-06
Add protection in use

dcmiddle:
Replace "encryption" to "cryptographic protection" to more accurately cover confidentiality and integrity.

Add techniques for protecting data during execution to complete the triad of in transit, at rest, and in use.

cleanup secretions.


lehors (approved):
I like it. Thanks @dcmiddle !

SecurityCRob (approved):
Excellent addition @dcmiddle , tyvm

process/security_baseline.md
```diff
@@ -231,6 +234,26 @@ Encrypting data in transit ensures that data transmitted over networks or betwee
   * Ed25519:
     * Fixed key length: 256 bits
 
+#### Data in Use
```
steiza, line 237: The content in this section is true, but I want to clarify the intention - is this section informational, or are we requiring TIs make use of confidential computing as part of security baseline / lifecycle maturation?

Providing this information to TIs is fine. I don't think we should require TIs make use of confidential computing in their lifecycle maturation.

dcmiddle: Thanks for the feedback. Can you say at what maturity level you think projects should protect data in use?

steiza: Well there's the capability ("protect data in use") and ways to implement that capability ("use confidential computing"). The way the document is structured, it seems like we're requiring TIs to use confidential computing in order to be graduated, which I don't agree with.

Maybe we could move this section to the end of the doc under its own heading, like "Informational" or "Additional Security Steps"?

steiza: Oops! I didn't realize the preview was collapsed, and this section is already in "Appendix" - never mind!

steiza (approved):

bobcallaway (approved):

PR #395, opened 2024-10-14
Update 2025 tac-and-scir-election-process.md

Naomi-Wash:


SecurityCRob (approved):
+1

lehors:
These changes don't look right to me. This is getting us back to the old process where the GB nominates people even before they know who's been elected when the whole point of the GB seats was that it would allow the GB to rebalance things if felt needed.

PR #398, opened 2024-10-15
Add security baseline reqs to project lifecycle templates

lehors:
Although we added security baseline requirements to our project lifecycle these requirements have not been added to the project lifecycle templates. This PR fixes that.

### DIFF
--- a/process/templates/PROJECT_NAME_graduation_stage.md
+++ b/process/templates/PROJECT_NAME_graduation_stage.md
@@ -41,7 +41,11 @@ When applicable, projects must have completed a security audit through a third p
 
 ### Security Baseline
 
-The project must meet the "[Security Baseline - To Become Graduated](https://github.com/ossf/tac/blob/main/process/security_baseline.md#security-baseline---to-become-graduated)" requirements.
+The project must meet all applicable Security Baseline requirements:
+ * [Security Baseline - Once Sandbox](https://github.com/ossf/tac/blob/main/process/security_baseline.md#security-baseline---once-sandbox)
+ * [Security Baseline - To Become Incubating](https://github.com/ossf/tac/blob/main/process/security_baseline.md#security-baseline---to-become-incubating)
+ * [Security Baseline - Once incubating](https://github.com/ossf/tac/blob/main/process/security_baseline.md#security-baseline---once-incubating)
+ * [Security Baseline - To Become Graduated](https://github.com/ossf/tac/blob/main/process/security_baseline.md#security-baseline---to-become-graduated)
 
 ### Project References
 The project must provide a list of existing resources with links to the repository, website, a roadmap, contributing guide, demos and walkthroughs, and any other material to showcase the existing breadth, maturity, and direction of the project.
--- a/process/templates/PROJECT_NAME_graduation_stage.md
+++ b/process/templates/PROJECT_NAME_graduation_stage.md
@@ -41,11 +41,11 @@ When applicable, projects must have completed a security audit through a third p
 
 ### Security Baseline
 
-The project must meet all applicable Security Baseline requirements:
- * [Security Baseline - Once Sandbox](https://github.com/ossf/tac/blob/main/process/security_baseline.md#security-baseline---once-sandbox)
- * [Security Baseline - To Become Incubating](https://github.com/ossf/tac/blob/main/process/security_baseline.md#security-baseline---to-become-incubating)
- * [Security Baseline - Once incubating](https://github.com/ossf/tac/blob/main/process/security_baseline.md#security-baseline---once-incubating)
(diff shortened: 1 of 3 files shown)


SecurityCRob (approved):
thanks for keeping all our collateral aligned!

sevansdell (approved):
lgtm

steiza (approved):

process/templates/PROJECT_NAME_incubation_stage.md
```diff
@@ -35,6 +35,10 @@ Project is integrated into the OpenSSF Scorecard
 When contributing an existing Project to the OpenSSF, the contribution must undergo license and IP due diligence by the Linux Foundation (LF). This step is only needed for the initial donation and only applicable here, if the project intends to join the OpenSSF Incubation stage.
   * "yes / no / not applicable. If yes, provide a link to the corresponding GitHub issue."
 
+### Security Baseline
+
+The project must meet the "[Security Baseline - To Become Incubating](https://github.com/ossf/tac/blob/308c777124a05f1903301400653f1a7a944bd7be/process/security_baseline.md#baseline---to-become-incubating)" requirements.
```
marcelamelara, line 40: There are also the "Once Sandbox" requirements that we probably want to add here as well: https://github.com/ossf/tac/blob/main/process/security_baseline.md#security-baseline---once-sandbox

process/templates/PROJECT_NAME_graduation_stage.md
```diff
@@ -39,6 +39,10 @@ Projects should harden their build systems in accordance with the SLSA Framework
 When applicable, projects must have completed a security audit through a third party and addressed audit findings and recommendations.
   * "link to results of security audit"
 
+### Security Baseline
+
+The project must meet the "[Security Baseline - To Become Graduated](https://github.com/ossf/tac/blob/main/process/security_baseline.md#security-baseline---to-become-graduated)" requirements.
```
marcelamelara, line 44: We will also want to check for the "Once Incubating" requirements here: https://github.com/ossf/tac/blob/main/process/security_baseline.md#security-baseline---once-incubating

marcelamelara (changes requested):
Thanks so much for these changes @lehors ! I do think we ought to add the "Once Sandbox/Incubating" requirements as well so TIs don't lose track of those.

process/templates/PROJECT_NAME_incubation_stage.md
lehors: Thanks for pointing this out. I think one could argue that the security baseline doc should be updated so that it is clear that the requirements for one stage include all of the previous ones but I expand the section here.

process/templates/PROJECT_NAME_graduation_stage.md
lehors: Actually I realized that this ought to also list the requirements applying to the previous stage since the baseline doc doesn't make that clear. I listed them all.

mlieberman85:
I'm still not a fan of how this is laid out. I think it's probably ok for now, but I think we should get some input from hands on keyboard maintainers of OpenSSF projects to make sure it's crystal clear to them who needs to do what and why.

One big thing is it's not clear whose responsibility is it to verify that these things are being done? There's a lot of talk around audit, but whose responsibility is it to audit? Whose responsibility is it to report findings?

lehors:
> I'm still not a fan of how this is laid out. I think it's probably ok for now, but I think we should get some input from hands on keyboard maintainers of OpenSSF projects to make sure it's crystal clear to them who needs to do what and why.
> 
> One big thing is it's not clear whose responsibility is it to verify that these things are being done? There's a lot of talk around audit, but whose responsibility is it to audit? Whose responsibility is it to report findings?

That's a fair point but that's really a comment on the security baseline, isn't it?
My expectation is that the project maintainers are responsible for implementing those requirements and the TAC is responsible for verifying that they have been when a project applies for a lifecycle change. But of course this leaves a lot in between and beyond the graduation stage.
One possibility is to add a prompt to the quarterly report template such as "List the security baseline requirements implemented in each of the projects related to your WG:".

If that's acceptable to you @mlieberman85 I suggest we open an issue to address this aside from this PR.

process/templates/PROJECT_NAME_incubation_stage.md
marcelamelara: Agreed. I think this was a discussion we had had with Dana as well. Seems like we should revisit this :)

marcelamelara (approved):
Thanks for the updates @lehors ! I agree with @mlieberman85 's comments about the layout, but don't think we should address that as part of this PR.

lehors:
@mlieberman85 @marcelamelara I added checkboxes which I think improves the layout better and prompts the applicant for their input.

PR #402, opened 2024-10-22
Add TAC decision type for TI quarterly updates

marcelamelara:
Since we're encouraging TIs to document their quarterly TAC updates via GitHub, this PR adds an explicit line to TAC-decision-process.md to document. 

This PR also removes the Blog decision since it doesn't apply to the TAC repo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # OpenSSF TAC Changelog
 
+* 2024-10-22: Add TAC decision type for TI quarterly updates [#402](https://github.com/ossf/tac/pull/402)
 * 2024-08-20: Label Security Metric project as archived in README
 * 2024-08-15: Updated TI-Gives+Gets.md to clarify Logo entitlement based upon lifecycle stage vs. stage Badge usage [#373](https://github.com/ossf/tac/pull/373)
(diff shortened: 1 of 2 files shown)


mlieberman85 (approved):
👍

lehors (approved):
I think one problem this document doesn't address is what we do in case the waiting period ends and we still don't have enough approvals. I'm not sure there is really a solution to this though, other than not re-electing TAC members who are failing to respond in a timely manner...

lehors:
> 5 approvals might be a little high? But I'm willing to give it a try.

I don't think it's high actually. We really should have about every TAC member sign off on all of these reports. This is how TAC members can stay informed about what's going on.

bobcallaway (approved):

sevansdell (approved):
I would like to see all TAC members have reviewed the TI quarterly update prior to the next TAC meeting. I think the 5 approvals you suggest is definitely doable, and I like the 7 day window because I would like to see TAC members in the repo at a weekly basis. Where it gets tricky is multiple people go on vacation at the time,  but I don't think that should happen too often.

SecurityCRob:
As an interested community member, I like the addition! Thanks !!

marcelamelara:
Thank you all! I also wasn't sure about the 5 approvals needed, but I agree with @lehors and @sevansdell that these reports is how we stay informed about the TIs. I say we try it out :)

Issue #430, opened 2025-01-08
TAC onboarding follow-ups


justaugustus:
These are follow-up from the "Open questions" sections:

- Are there additional access requirements not accounted for in this document?
- What are our moderation responsibilities and privileges (GitHub, mailing lists, Zoom)?
- How do we manage rescheduling the TAC meeting to accommodate timezones and/or new member obligations?
- Explicitly, what are the responsibilities of the TAC Chair and Vice Chair?

I'll clean up this list as we move through the items in https://github.com/ossf/tac/issues/426.

PR #429, opened 2025-01-04
process: Add initial onboarding document for incoming TAC members

justaugustus:
Part of https://github.com/ossf/tac/issues/426.
cc: @ossf/tac

PR #432, opened 2025-01-09
Simplify docs around Chair election, roles, and responsibilities

justaugustus:
This is a follow-up to @marcelamelara's [review comment](https://github.com/ossf/tac/pull/429#pullrequestreview-2532703309) around simplification of references to TAC Chair roles, responsibilities, and election process.

ref: https://github.com/ossf/tac/issues/430


elections/tac-and-scir-election-process.md
```diff
@@ -73,7 +73,15 @@ The Governing Board will appoint three people from the list of TAC and SCIR Nomi
 
 ## TAC Chair and Vice Chair Election
 
-After the results are announced the TAC members will elect a Chair and a Vice Chair with 1 week for nominations and 1 week for votes. Results may be announced when all votes are cast. The candidate with the most votes will serve as Chair and the second as Vice Chair. The Chair may however choose to swap seat with the Vice Chair if agreed by the Vice Chair.
+After the results of the community elections and Governing Board appointments are announced, the TAC members will elect a [Chair and a Vice Chair](/process/tac-member-R&Rs.md#tac-chair).
+
+The TAC Chair nomination period shall commence as soon as new TAC members are seated (typically, at the beginning of the year) and shall conclude at the end of the first TAC meeting of the year.
```
justaugustus, line 78: @ossf/tac — I've edited this section to reflect what we're currently in the process of doing, as opposed to the one week for nominations and one week for votes.

Happy to revert that piece, but calling it out, as it's a decision point that merits discussion.

lehors: This is fine with me but I have to say that we've got to do better at following whatever process we have. This hasn't been the case unfortunately. We just can't keep changing our documentation to match what we did just because we ignored our documented process. It's fine to change our process of course but it's got to be intentional.

lehors (approved):

elections/tac-and-scir-election-process.md
marcelamelara: We still did take ~1 week for nominations (it was relatively informal) prior to the Jan 7 TAC meeting, and are taking 1 week for voting. So, I actually don't think that's changed. And we won't be concluding this year's election by the first TAC meeting, but I'm ok with bounding the timeframe for the election to the second TAC meeting. The biggest change I see is the involvement of OSSF staff this year. In the past, voting was done "internally" within the TAC.

justaugustus: > And we won't be concluding this year's election by the first TAC meeting [...]

@marcelamelara — just a point of clarification, the proposed text says that:

> The TAC Chair _nomination_ period [...] shall conclude at the end of the first TAC meeting of the year.

i.e., nominations need to be in by the end of the first meeting so that we can complete elections ahead of the second meeting

I'll refine the language a bit more based on the other feedback.

marcelamelara: Ah! Thanks for clarifying, I misread that. I'm do worry that bounding nominations to the first TAC meeting may end up giving the new TAC only a couple of days (depending on the meeting schedule), but I'm generally still ok with saying that the TAC _SHOULD_ aim to complete the chair/vice-chair election process by the second TAC meeting of the year.

elections/tac-and-scir-election-process.md
```diff
@@ -73,7 +73,15 @@ The Governing Board will appoint three people from the list of TAC and SCIR Nomi
...
+
+OpenSSF staff will coordinate the election process and results will be announced soon after all votes have been cast.
```
gkunz, line 80: I think we should add (back) a bit more detail and a concrete timeline to this aspect of the process. For instance, stating that the election period is 1 week, starting after the conclusion of the nomination phase / the end of the first TAC meeting. This adds clarity to the process (helps staff) and is in fact in alignment with the currently documented process and what we decided to do this year.

gkunz: In line with my comment above, I suggest to add a concrete timeline for the voting period. The suggestion here is 1 week, as it is also currently documented.

```suggestion
The voting period is 1 week, starting immediately after the conclusion of the nomination period. OpenSSF staff will coordinate the election process and results will be announced soon after all votes have been cast.
```

elections/tac-and-scir-election-process.md
steiza, line 78:
```suggestion
The TAC Chair nomination period shall commence as soon as new TAC members are elected (typically, around the beginning of the year) and shall conclude at the end of the first TAC meeting of the year.
```

Minor: to me, "seated" means when the new TAC meets for the first meeting of the year, whereas nominations can start as soon as the election results are announced.

process/tac-member-R&Rs.md
```diff
@@ -2,23 +2,51 @@
...
+### Responsibilities
+
+- Fulfill TAC member responsibilities
+- Fulfill the responsibilities of the TAC Chair's [additional roles](#additional-chair-roles)
+- Preside over meetings
+- Set the agenda for meetings
+- Ensure meeting minutes are taken
+- Coordinate the TAC to conduct technical management and oversight of OpenSSF Technical Initiatives
+- Organize the selection of a TAC member to serve on the OpenSSF [Governance Committee][gov-comm], *in addition to the TAC Chair's membership*
+
```
steiza, line 46:
```suggestion
- Give TAC updates to the Governing Board as requested, including committees of the Governing Board

```

steiza (commented):
Thank for putting this together! I have a couple of minor questions, and it looks like maybe there's some other comments / suggested changes to resolve before a final review?

process/tac-member-R&Rs.md
```diff
@@ -2,23 +2,52 @@
...
+### Responsibilities
+
+- Fulfill TAC member responsibilities
+- Fulfill the responsibilities of the TAC Chair's [additional roles](#additional-chair-roles)
+- Preside over meetings
+- Set the agenda for meetings
+- Ensure meeting minutes are taken
+- Coordinate the TAC to conduct technical management and oversight of OpenSSF Technical Initiatives
+- Organize the selection of a TAC member to serve on the OpenSSF [Governance Committee][gov-comm], *in addition to the TAC Chair's membership*
```
justaugustus, line 45:
```suggestion
```

Removing this, following a discussion in the private channel.
i.e., 

> The selection process of the TAC Members will be determined by the TAC (via a vote if necessary). However one TAC Member out of the 2 must be the TAC Chair (the TAC Chair seat does not count towards the Governing Boards seven seats).

ref: https://github.com/ossf/Governance-Committee/blob/main/CHARTER.md#membership

justaugustus:
@ossf/tac — I've processed the review feedback and this should be ready to merge now.

steiza (approved):
Looks great - thanks!

gkunz (approved):
Thank you driving this

lehors (approved):
At this point there are more lines being added than deleted so I'm not sure the title is quite right - it probably should say something like "cleanup docs ..." - but that's a detail that I'm only pointing out because I tend to be pedantic about this kind of stuff. ;-)
It LGTM otherwise.
Thanks!

mlieberman85 (approved):
LGTM!

marcelamelara (approved):
Thanks @justaugustus ! LGTM

PR #441, opened 2025-01-22
Updates links in README.md projects table

GeauxJD:
Changes include: updating outdated Meeting Notes links for projects to 2025 docs, adding Minder links, removing Security Metrics from archived Metrics WG and updating Zarf github link.


marcelamelara (approved):
Thanks @GeauxJD !

lehors (commented):
LGTM but I have to say that I don't think we should have that level of information in the TAC README. Instead we should ensure that every TI has that info on their home page, easy to find and only have a link to each TI's home page in the TAC README.
Duplicating information increases the maintenance burden and leads to discrepancies.

GeauxJD:
I closed this by mistake but will make a new PR that reduces the information/links as Arnaud suggests above.

PR #490, opened 2025-06-09
Create 2025 Q2 Sigstore Summary

Hayden-IO:


gkunz (approved):
great work. Thank you!

bobcallaway (approved):

steiza (approved):
Thanks!

lehors:
The report is great. I just have a couple of questions out of curiosity. Apologies for my ignorance.
1. How much is the client impacted by the change in the transparency log implementation? I would have (naively) thought this was "an implementation detail" within the server that isn't visible to the client. From what I read that's not the case, though.
2. How is the transition from the old log to the new one handled? I don't expect you to be able to move the old data to the new implementation so how do you handle continuity?
Thank you, and feel free to point me to relevant documentation that answers those questions if something succinct is already out there.

lehors (approved):

PR #341, opened 2024-06-05
add zarf as an openssf sandbox project

salaxander:

PR #497, opened 2025-07-02
chore(docs): update zarf documentation to reflect current state

brandtkeller:
I had some discussions and believed the outcomes were that this should be a living document. Wanted to update references and list of current maintainers to be accurate. 

Open to feedback on any items that should/should-not be updated as well.

### DIFF
--- a/process/project-lifecycle-documents/zarf_sandbox_stage.md
+++ b/process/project-lifecycle-documents/zarf_sandbox_stage.md
@@ -2,9 +2,9 @@
 
 ### List of project maintainers
 The project must have a minimum of three maintainers with a minimum of two different organizational affiliations.
-  * Sarah Christoff, Defense Unicorns, @sscristoff-du
+  * Kit Patella, Defense Unicorns, @mkcp
   * Austin Abro, Defense Unicorns, @AustinAbro321
-  * Lucas Rodriguez, Defense Unicorns, @lucasrod16
+  * Brandt Keller, Defense Unicorns, @brandtkeller
   * Danny Gershman, Radius Method, @dgershman
 
 ### Sponsor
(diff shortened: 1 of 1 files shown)


marcelamelara (approved):
Thanks for sending this @brandtkeller !

justaugustus (commented):
> I had some discussions and believed the outcomes were that this should be a living document.

@brandtkeller — Did anyone from TAC weigh in on these discussions?

IMO, the project's documentation should reflect the current state of affairs (maintainers, links, etc.), with the maturity model application representing the project at the time of its submission (https://github.com/ossf/tac/pull/341).

It's probably fine to update the links, but I would leave the maintainers section as-is.

Curious to hear thoughts from other @ossf/tac members.

brandtkeller:
> @brandtkeller — Did anyone from TAC weigh in on these discussions?

The discussions were with LF/OSSF staff but I would also love TAC comments. 

My interest was the source of truth for extending invitations to respond as a project. We've seen prior maintainers being contacted and my interest in updating if this was _a_ source of truth.

justaugustus:
> My interest was the source of truth for extending invitations to respond as a project. We've seen prior maintainers being contacted and my interest in updating if this was _a_ source of truth.

I would treat the [Projects listing](https://github.com/ossf/tac?tab=readme-ov-file#projects) as the source of truth, but given that table links out to maturity applications, I can see that being confusing for folks.

---

Procedural note for TAC/Staff:
Here's how Kubernetes handles this: https://github.com/kubernetes/community

lehors:
I don't see this document being a living doc at all and don't think the list of maintainers should be updated.
It might make sense to fix a broken link but this should largely be left unchanged because it is meant to show the state of the project at the time of application for the correspondront lifecycle stage. If you kept updating it at some point you would end up with an application for sandbox which shows a protect qualifying for, say, graduation.

PR #498, opened 2025-07-02
Add Q3 update for the AI/ML WG.

mihaimaruseac:


TI-reports/2025/2025-Q3-AI-ML-WG.md
```diff
@@ -0,0 +1,117 @@
...
+that _there is not enough skilled developer time for deep security work_. In
+particular, for dependencies of OSS projects, this results in the need to have a
+team of "firefighters"/"fire prevention squad". These cannot be part-time.
+
+Next, the AI economics SIG also looked at the impact of AI slop on vulnerability
+management and how this can result in bogus security reports being sent to
+maintainers, adding to their stress.
+
+The AI economics SIG and the AI/ML working group at large will continue working
+on these issues.
```
steiza, line 58: I read https://github.com/ossf/ai-ml-security/issues/27, but the objective/scope of the AI economics SIG is a little unclear to me. That issue describes a sort of expected-loss version of https://github.com/ossf/criticality_score, and while I have questions about the methodology, I can see where the term "economics" comes into play there.

But what's the economic tie-in with SBOMs (AI or otherwise)? FAIR is from risk management / compliance frameworks, but if anything is more actuarial than economic? I also love looking for ways to decrease AI slop vulnerability reports, but I keep tripping over the term "economics".

What's the theme that ties these workstreams together? Is there a more apt term than "economics"?

steiza (approved):
Really exciting updates on model signing and the AI DevSecOps whitepaper. There's some interesting ideas listed under the AI Economics SIG, but the scope and how we're tying things back to the label "economics" to is a little unclear to me. Regardless, thank you for this great update!

TI-reports/2025/2025-Q3-AI-ML-WG.md
mihaimaruseac: CC @0xAverageUser

0xAverageUser: @steiza @mihaimaruseac 

> Is there a more apt term than "economics"?

That's an interesting question. We use terms like cost, price, utility, value, risk in the context of open source security so enterprise finance and business teams, as well as public sector policy makers understand open source risk. The terminology and machinery is from economics, but we could also use "finance". 

> I read [ossf/ai-ml-security#27](https://github.com/ossf/ai-ml-security/issues/27), but the objective/scope of the AI economics SIG is a little unclear to me. 

The objective of the SIG can be summarized as follows: predict which open-source packages, if compromised, would cause the highest financial losses for Fortune 500 companies.

ossf/criticality_score is a great contribution to identifying high cyber-risk packages, but in FAIR's terms the "loss magnitude" is missing, so that's only half of the equation, at most. 

> But what's the economic tie-in with SBOMs (AI or otherwise)? 

There's the linux census but without (anonymized) SBOMs it's not clear how large enough ecosystem datasets could be aggregated from large enterprises' proprietary code automatically. On the level of the individual enterprise, the program goal is to upgrade SBOMs to become a "software balance sheet" - that means analyzing the continuous benefits but also maintenance cost and risk from adopted open source.

> What's the theme that ties these workstreams together? 

That's also a good question and I agree this isn't sharp. [The SIG's perspective on AI Slop](https://docsend.com/view/u73ugdeg8vrxqbe5) is to focus on the effects of reduced cost of fake vulnerability reports, which shifts the economic equilibrium of bug bounties, vulnerability management programs, the CVE system in the favor of attackers, because the cost of triage and verification has not decreased on a similar level. 

The cost of attack versus mitigation cost is an overlooked but significant element of Security at Scale. Security Economics is the most apt term for what ties the workstreams we came up with so far, but definitely open to a good suggestion, maybe from a native English speaker.

steiza: > The cost of attack versus mitigation cost is an overlooked but significant element of Security at Scale. Security Economics is the most apt term for what ties the workstreams we came up with so far, but definitely open to a good suggestion, maybe from a native English speaker.

Maybe AI risk management SIG? Or AI risk mitigation SIG? Either of those could both refer to new risks associated with when you use AI, as well as the risks when other people use AI (like in the case of AI slop vulnerability reports).

Definitely open to other people's suggestions here!

lehors (approved):

TI-reports/2025/2025-Q3-AI-ML-WG.md
```diff
@@ -0,0 +1,117 @@
...
+need to work on the documentation PRs for this, as well as on the documentation
+PRs for formalizing the sandbox stage for the AI Economics SIG.
+
+For model signing, we are looking at more integrations and demos. We are talking
+with various companies (HuggingFace, Cohere, Dell, RedHat), directly or via
+CoSAI. We have given demos at OSS NA, devconf.cz and academic conferences.
+
+The model signing project is also looking at incorporating signatures generated
+on GPUs. We are also looking at merging with the work from the Atlas project
+from Intel.
```
marcelamelara: Really appreciate the shoutout! Adding a link for more context.
```suggestion
on GPUs. We are also looking at merging with the work from the [Atlas project](https://sched.co/21A0r)
from Intel.
```

TI-reports/2025/2025-Q3-AI-ML-WG.md
```diff
@@ -0,0 +1,117 @@
...
+
+Regarding model signing, we also had a presentation from Purdue about generating
+model signatures directly on GPU and validating them on GPU at the user's site.
+
+### Up Next
+
+We started with
+[an issue to ask the community for ideas to work on in 2025](https://github.com/ossf/ai-ml-security/issues/26),
+and now we also have a
+[new long term plan](https://docs.google.com/document/d/1I-GDKV3rL3jPwPK3xO5hKh8feGBouTwTUY3aqj0-0IY/edit?pli=1&tab=t.0).
```
marcelamelara, line 95: This sort of plan was very much needed, so I'm really happy to see the doc. My remaining question / feedback is: As the WG progresses through this plan, can the WG refine its scope to more clearly differentiate from other LF/ecosystem efforts in AI? For example, one of the roles I can see the AI/ML WG taking on is to host ML model/lifecycle security tooling.

TI-reports/2025/2025-Q3-AI-ML-WG.md
marcelamelara: I've seen several objective statements at this point:

From the [original proposal](https://github.com/ossf/ai-ml-security/issues/27):
> this SIG will deliver practical guidance and tools to assist OSS consumers and stakeholders in assessing and optimizing their security investments.

From @0xAverageUser response:
> predict which open-source packages, if compromised, would cause the highest financial losses for Fortune 500 companies.

> The SIG's perspective on AI Slop is to focus on the effects of reduced cost of fake vulnerability reports, which shifts the economic equilibrium of bug bounties, vulnerability management programs, the CVE system in the favor of attackers, because the cost of triage and verification has not decreased on a similar level.

Here's how I interpret what these objectives are about:
The first: Where and how much are stakeholders (this is itself broad) investing in OSS security, and where _should_ stakeholders invest more/less in OSS security?
The second: Which OSS compromises are most expensive to fix?
The third: What is the added cost of having to detect and mitigate compromises overlooked through AI slop?

These are all important, but my sense is that more clearly and precisely articulating the vision and roadmap of the SIG will help with the current scoping issues because the answer to some of these questions depends on getting answers to others. There's also the added layer to help answer some of these questions using AI, hence the overly broad SIG name.

My recommendation: I agree with @steiza that more precision in the name of the SIG would be helpful, but also think that the approach of using AI to do mitigation cost analysis of AI-generated security metadata is muddying the objectives. This is not to say that this isn't the reality we live in, but I think the SIG will possibly have an easier time planning and articulating its roadmap if there is a clearer split between the "AI for security" vs. "security for AI" pieces, and to focus on one of these sides of the coin first, then the other.

marcelamelara (approved):
Thanks for this update @mihaimaruseac !

bobcallaway (approved):

PR #517, opened 2025-09-03
Update tac-and-scir-election-process.md

kj-powell:
Updated election timelines for 2025

### DIFF
--- a/elections/tac-and-scir-election-process.md
+++ b/elections/tac-and-scir-election-process.md
@@ -106,12 +106,12 @@ About 10 months after the previous cycle, a committee of Election Officials will
 
 | Date | Action |
 | :--- | :--- |
-| October 14 | Self-Nomination and Voter registration period starts. Send call for nominees and eligible voters in email and on Slack.|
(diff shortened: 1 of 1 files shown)


gkunz (approved):

elections/tac-and-scir-election-process.md
```diff
@@ -106,12 +106,12 @@ About 10 months after the previous cycle, a committee of Election Officials will
```
steiza, line 106: Minor: do we want to change "2025 Timeline" above?

Last year, we used that to mean "the election for the 2025 TAC". This year we're using it to mean "the election that will take place in 2025". 🤷 Maybe it doesn't matter.

kj-powell: Yes let's change it so it stays consistent. Also, I realized that one of the dates is wrong (should be Oct 27 where it says Oct 30).

justaugustus: Fixed in https://github.com/ossf/tac/pull/518.

PR #585, opened 2026-03-18
Add TAC leave of absence guidance

marcelamelara:
Per the discussion during the [3/3/2026 TAC meeting](https://docs.google.com/document/d/15cW5oqu5Jp1iRdDNIrcUnD5gcoKxPKkJ0bc9qjzf5h0/edit?tab=t.0), this PR adds the TAC member leave of absence policy.

Please note that the voting threshold in `gitvote.yml` is already set to a simple majority (55%), so an adjustment may not be needed.


gkunz (approved):
Looks good to me. I like the pragmatic approach.

PR #593, opened 2026-03-31
Transition Malicious Packages under Securing Software Repositories WG

kborchers:
This will effectively move the Malicious Packages project from the Securing Critical Projects WG to the Securing Software Repositories WG.


steiza (approved):
This is what we discussed at the last Securing Repos WG meeting! cc @miketheman

PR #602, opened 2026-04-26
Fixed a level error to be coherent

JustinCappos:
Since you're referring to Incubating benefits here, I think this must be saying greater than Sandbox instead of Incubating.


justaugustus (approved):
Thanks for the catch, @JustinCappos!

PR #656, opened 2026-09-16
Ortelius Sandbox Proposal

TracyRagan:
This pull request proposes Ortelius as an OpenSSF Sandbox Project under the Vulnerability Disclosures Working Group. Ortelius extends vulnerability disclosure into post-deployment operations by correlating SBOMs, software releases, dependencies, deployment evidence, and endpoint information with continuously updated vulnerability data. The project helps users determine whether newly disclosed vulnerabilities affect software they are actually running, where impacted components are deployed, and which findings require remediation. This proposal aligns Ortelius with the Vulnerability Disclosures Working Group’s efforts to improve how vulnerability information is communicated, consumed, and acted upon, while providing an open source implementation focused on the operational gap between vulnerability disclosure and remediation.

PR #657, opened 2026-09-16
Update README.md

TracyRagan:
Added proposed Ortelius Project under the Vulnerability Disclosures WG


justaugustus:
@TracyRagan — Can you move this README update into https://github.com/ossf/tac/pull/656?

That way if/when it's approved all of the updates happen in a single place (instead of this accidentally getting merged out-of-band).

